PR #1700, opened 2025-02-25
🐞 Admin Portal > Reports > Timezone shown is not showing timezone

yuvalkom-M:
parent issue: https://github.com/sequentech/meta/issues/6191


hasura/metadata/actions.graphql
```diff
@@ -31,9 +31,9 @@ type Mutation {
     create_ballot_receipt(
         ballot_id: String!
         ballot_tracker_url: String!
-        tenant_id: uuid!
```
xalsina-sequent: Should tenant id be removed from this call?

hasura/metadata/actions.graphql
```diff
@@ -31,9 +31,9 @@ type Mutation {
...
         election_event_id: uuid!
         election_id: uuid!
+        user_timezone: String!
```
xalsina-sequent: Make this parameter optional so it can be used by old api integrations

hasura/metadata/actions.graphql
```diff
@@ -244,6 +245,7 @@ type Mutation {
         report_mode: String!
         tenant_id: String!
         election_event_id: String
+        user_timezone: String!
```
xalsina-sequent: Make this parameter optional so it can be used by old api integrations

packages/harvest/src/routes/create_ballot_receipt.rs
```diff
@@ -6,25 +6,23 @@ use crate::services::authorization::authorize_voter_election;
...
-    pub ballot_tracker_url: String,
-    pub election_event_id: String,
-    pub election_id: String,
-    pub time_zone: Option<TimeZone>,
-    pub date_format: Option<DateFormat>,
+    ballot_id: String,
+    ballot_tracker_url: String,
+    election_event_id: String,
+    election_id: String,
+    user_timezone: String,
```
xalsina-sequent: This should be optional

PR #1855, opened 2025-05-29
✨ Add versioning to election event configs

omri81:
Parent issue: https://github.com/sequentech/meta/issues/6333


packages/windmill/src/services/import/import_election_event.rs
```diff
@@ -445,13 +399,54 @@ pub async fn decrypt_document(
     Ok(temp_file_path)
 }
 
+fn extract_major(version: &str) -> Option<u32> {
```
xalsina-sequent: This could give an error if the version string includes a 'v' character like v1.0.0.

PR #2095, opened 2025-10-23
✨ Tally > Ballot Images: add election_alias

yuvalkom-M:
parent issue: https://github.com/sequentech/meta/issues/8680

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -2,8 +2,8 @@
 # ── 1. pin rust-overlay ──────────────────────────────────────────────
 let
   rustOverlay = import (builtins.fetchTarball {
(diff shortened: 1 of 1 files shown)


BelSequent:
image

There is an issue: The alias is also displayed as a name, it seems that the name gets overwritten with the alias somewhere. I followed these steps:
- Initially the election has no alias so the election name is shown in the ballot image.
- Modified the template to show the election_alias variable under the election name:
image

- Added "Presidental Alias" as an alias to the election.
- Published changes and run tally to download the ballot images.
- Then the election alias is shown but also the election name is overwritten somehow with the alias.

BelSequent (changes requested):
Needs work

BelSequent (approved):

PR #2141, opened 2025-11-03
✨ Delegate voting with imports 

xalsina-sequent:
Parent issue: https://github.com/sequentech/meta/issues/7683


packages/windmill/src/services/ceremonies/insert_ballots.rs
```diff
@@ -208,13 +214,16 @@ pub async fn insert_ballots_messages(
                     let users_join_idexes = 0;
                     let contest_id = tally_session_contest.contest_id.clone();
 
+                    let delegate_count_index = if is_delegated { Some(1) } else { None };
```
yuvalkom-M: Maybe just add a comment explaining what the index '1' means. This will make it clear that if the voters csv file changes, this number may needs to be updated as well.

PR #2148, opened 2025-11-05
✨ Support public/unencrypted plaintext vote/voting

xalsina-sequent:
Parent issue: https://github.com/sequentech/meta/issues/838

### DIFF
--- a/.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5.json
+++ b/.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5.json
@@ -1786,6 +1786,24 @@
         "clientRole": false,
         "containerId": "9c67adb8-ae12-43ad-a3f4-2bc275fc888a",
         "attributes": {}
+      },
+      {
+        "id": "5640710d-b40b-44cb-888f-f6b7277efedd",
+        "name": "generate-preview",
+        "description": "",
+        "composite": false,
+        "clientRole": false,
+        "containerId": "50ef5e3e-42c6-4588-b46a-bb938aa8d849",
+        "attributes": {}
+      },
+      {
+        "id": "bddb99dc-bf6d-44a9-8f0a-d4830ce1b92d",
+        "name": "preview-read",
+        "description": "",
+        "composite": false,
+        "clientRole": false,
+        "containerId": "50ef5e3e-42c6-4588-b46a-bb938aa8d849",
+        "attributes": {}
       }
     ],
     "client": {
@@ -2373,7 +2391,9 @@
         "election-dashboard-tab",
         "voter-voted-edit",
         "voter-email-tlf-edit",
-        "transmission-report-generate"
+        "transmission-report-generate",
+        "preview-read",
+        "generate-preview"
       ],
       "clientRoles": {}
     },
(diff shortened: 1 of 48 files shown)


packages/voting-portal/src/routes/ReviewScreen.tsx
```diff
@@ -398,11 +402,15 @@ const ActionButtons: React.FC<ActionButtonProps> = ({
             }
         }
 
-        let hashableBallot: IHashableSingleBallot | IHashableMultiBallot | undefined
```
Findeton: Interesting, I need to review this change in further detail.

packages/voting-portal/src/services/BallotStrategy.ts
```diff
@@ -0,0 +1,130 @@
...
+    service: IBallotService
+): IBallotStrategy => {
+    switch (policy) {
+        case EElectionEventContestEncryptionPolicy.SINGLE_CONTEST:
+            return new SingleContestStrategy(service)
+        case EElectionEventContestEncryptionPolicy.MULTIPLE_CONTESTS:
+            return new MultiContestStrategy(service)
+        case EElectionEventContestEncryptionPolicy.PLAINTEXT:
+            return new PlaintextStrategy(service)
+        default:
```
Findeton: so in packages/sequent-core/src/ballot.rs we have:

pub contest_encryption_policy: Option<ContestEncryptionPolicy>,

which is optional even though  ContestEncryptionPolicy has a default. This means that here by default we'll get an error.

packages/sequent-core/src/plaintext_ballot.rs
```diff
@@ -0,0 +1,393 @@
...
+
+#[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Debug, Clone)]
+pub struct RawHashablePlaintextBallot<C: Ctx> {
+    pub version: u32,
+    pub issue_date: String,
+    pub contests: Vec<HashablePlaintextBallotContest<C>>,
+}
+
+impl AuditablePlaintextBallot {
+    pub fn deserialize_contests<C: Ctx>(
```
Findeton: We should add tests for serializing/deserializing for the whole file.

Findeton (changes requested):
Great work Xavier! I added some comments, please address them. Basically adding some tests, removing a magic string and adding documentation.

.devcontainer/scripts/build-sequent-core.sh
```diff
@@ -9,7 +9,8 @@ set -euo pipefail
 # flag, which is not supported when compiling C code for WebAssembly targets
 # (used by the ring cryptographic library):
 export NIX_HARDENING_ENABLE=""
-export CFLAGS_wasm32_unknown_unknown="-O3 -ffunction-sections -fdata-sections -fno-exceptions"; 
+# Append optimization flags to existing CFLAGS (preserving include paths from flake.nix)
+export CFLAGS_wasm32_unknown_unknown="${CFLAGS_wasm32_unknown_unknown} -O3 -ffunction-sections -fdata-sections -fno-exceptions"
```
Findeton: what does this do?

packages/windmill/src/tasks/execute_tally_session.rs
```diff
@@ -761,6 +803,48 @@ async fn map_plaintext_data(
     let messages: Vec<Message> = protocol_manager::convert_board_messages(&board_messages)?;
     print_messages(&messages, &bulletin_board)?;
 
+    // Since we skipped key ceremony for plaintext we generate the configuration using dummy trustees.
```
Findeton: Why do we need to generate dummy keys here if it's an unencrypted election event?

PR #2198, opened 2025-12-01
✨ Browser-based trustees

Findeton:
Parent issue: https://github.com/sequentech/meta/issues/9536


.devcontainer/docker-compose-base.yml
```diff
@@ -521,17 +521,18 @@ services:
     build:
       context: ../packages/
       dockerfile: ./braid/Dockerfile
+    #entrypoint: ["/bin/sh", "-c", "while true; do sleep 3600; done"]
```
BelSequent: Leftover?

packages/wbraid/build-wasm.sh
```diff
@@ -0,0 +1,105 @@
+# SPDX-FileCopyrightText: 2025 Sequent Tech Inc <legal@sequentech.io>
```
BelSequent: I think we should move this script to braid and remove wbraid. But first check if there is anything else worth keeping in this folder.

BelSequent (changes requested):
Apart from these things I comment, the sequent-core wasm build GAction is failing BUT I have fixed it in the follow up branch so you can ignore this for now.
I did not review all the files yet, but I think it´s a good start. There are also some conflicts to fix with main.

BelSequent (approved):

PR #2298, opened 2025-12-28
🐞 Voting Portal > Candidate images broken after export then import the event

yuvalkom-M:
https://github.com/sequentech/meta/issues/9087


yuvalkom-M:
No need

PR #2525, opened 2026-03-18

yuvalkom-M:
parent issue: https://github.com/sequentech/meta/issues/11566


Findeton (changes requested):
- lib.rs:5 adds a crate-wide allow for missing_docs while Cargo.toml:167 and Cargo.toml:175 introduce deny rules for missing docs. Remove the crate-wide allow, and instead of using missing_docs, add the docs (AI can help with it)
- Use cargo clippy (for crates where it should pass, in this case just sequent core for now) in the vs code task for linting
- Add documentation on how to use cargo clippy and what is the linting config
- Add a GH action for clippy linting (for specific crates).
- Cargo clippy should be used with `--all-features`.
- Modify the GH action for docs so that cargo docs (for all crates) are generated and pushed (copied), so that the preview and the "main" branch include the documentation generated by cargo doc.
- Related to the previous comment, I pushed a small commit to improve the generated cargo doc (so that all features are included).

packages/sequent-core/src/services/area_tree.rs
```diff
@@ -259,25 +272,27 @@ where
...
-    // area-contests. Note that this will include
-    // indirect/inherited ones.
+    /// For a given `TreeNode` of type `ContestsData`, return all
+    /// area-contests. Note that this will include indirect/inherited ones.
+    #[must_use]
+    #[allow(clippy::used_underscore_binding)]
     pub fn get_contest_matches(
         &self,
-        contest_ids: &HashSet<String>,
+        _contest_ids: &HashSet<String>,
```
eselimsen: This parameter is used, right? I think we should remove the `#[allow(clippy::used_underscore_binding)]` above, and use it without `_`

packages/sequent-core/src/ballot_codec/bigint.rs
```diff
@@ -1,72 +1,107 @@
...
-    let mut i = choices.len() - 1;
-    while 0 == choices[i] {
-        i -= 1;
+    if choices.is_empty() {
+        return RawBallotContest { bases, choices };
     }
+    let i = choices
+        .iter()
+        .rposition(|&choice| choice != 0)
+        .unwrap_or(choices.len().saturating_sub(1));
```
eselimsen: I think this changes the logic a bit. In the previous version, if the input is full of zeroes, it'd panic. The new version doesn't panic, but would an element anyway. I think `if let` would be more predictable, like so:
```rs
if let Some(i) = choices.iter().rposition(|&choice| choice != 0) {
  choices.remove(i);
  bases.remove(i);
}
```

packages/sequent-core/src/services/connection.rs
```diff
@@ -236,6 +250,7 @@ async fn read_access_token(
     };
 
     if let Some(data) = token_resp_ext_opt {
+        #[allow(clippy::arithmetic_side_effects, clippy::cast_possible_wrap)]
```
eselimsen: Any reason to not apply the same safety checks here?
🤔

packages/sequent-core/src/services/pdf.rs
```diff
@@ -573,31 +612,40 @@ impl PdfRenderer {
...
             s3::get_file_from_s3(bucket, output_filename)
                 .await
                 .map_err(|err| anyhow!("could not retrieve file from S3: {err:?}"))
         }
     } else {
+        /// Returns None for private S3 bucket when S3 is not enabled.
         fn s3_private_bucket() -> Option<String> {
             None
         }
-        fn s3_bucket_path(path: String) -> Option<String> {
```
eselimsen, line 591: I think it would be better to retain `Option` in this case, is it infallible?

packages/sequent-core/Cargo.toml
```diff
@@ -154,3 +157,36 @@ sqlite = ["dep:tokio", "dep:rusqlite"]
...
+[lints.rust]
+missing_docs = "deny"
+unsafe_code = "forbid"
+private_interfaces = "warn"
+private_bounds = "warn"
+unnameable_types = "warn"
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage,coverage_nightly)'] }
+
+[lints.clippy]
+missing_docs_in_private_items = "deny"
```
eselimsen, line 175: I think we should consider adding `allow_attributes_without_reason` here. There are inconsistent `allow` usages, breaking their point, be it for except, future sizes, arithmetic or docs.

README.md
```diff
@@ -390,3 +390,117 @@ when running in the terminal `devenv shell` or in the `Rebuild Container` vscode
...
+Explanation
+--no-deps: Lints only this workspace (skips dependencies)
+--all-features: Ensures all feature-gated code is checked
+-- -A warnings: Allows warnings (only errors will fail)
+
+- Note: This is convenient for development, but CI should be stricter.
+
+### Lint Configuration
+
+Lint rules are defined in Cargo.toml under [lints.*].
```
eselimsen: Do we need to copy the public lint descriptions in our readme as well?
🤔

packages/sequent-core/src/util/date_time.rs
```diff
@@ -41,9 +46,13 @@ pub fn generate_timestamp(
...
         TimeZone::UTC => now.format(&date_format).to_string(),
         TimeZone::Offset(offset) => {
-            let duration = Duration::hours(offset as i64);
-            let fixed_offset =
-                FixedOffset::east_opt(duration.num_seconds() as i32);
+            let duration = Duration::hours(i64::from(offset));
+            let fixed_offset = FixedOffset::east_opt(
+                match i32::try_from(duration.num_seconds()) {
+                    Ok(secs) => secs,
+                    Err(_) => return now.format(&date_format).to_string(),
```
eselimsen: This unlikely to fail considering seconds with i32 for offsets, but I think it is better to add a warning or error log here at least, just to avoid failing silently.

packages/sequent-core/src/ballot.rs
```diff
@@ -963,16 +1092,24 @@ pub struct ElectionEventLanguageConf {
     Clone,
     Default,
 )]
+/// Presentation configuration for an election event.
+#[allow(missing_docs)]
 pub struct ElectionEventPresentation {
+    /// Internationalized content for the event.
     pub i18n: Option<I18nContent<I18nContent<Option<String>>>>,
-    pub materials: Option<ElectionEventMaterials>,
+    pub activated: Option<ElectionEventMaterials>,
```
eselimsen: Is this exposed through the API, would it break the frontend?

packages/sequent-core/src/fixtures/ballot_codec.rs
```diff
@@ -220,115 +233,19 @@ fn get_contest_plurality() -> Contest {
...
 }
 
-fn get_contest_irv() -> Contest {
-    let mut contest = get_contest_plurality();
-    contest.counting_algorithm = Some(CountingAlgType::InstantRunoff);
-    contest.max_votes = 3;
-    contest
-}
-
-pub fn get_irv_fixture_valid_ballot() -> BallotCodecFixture {
```
eselimsen, line 238: Why was this removed, it breaks `test_irv_invalid_ballot` test for me, does it run for you?
🤔

eselimsen (changes requested):
Agree with secret leak comments from Copilot. I think we should use a `Secret` newtype wrapper to protect them, and reveal explicitly only when required.
I think we need consistent `except` and arithmetic handling, there are liberal `#[allow]` usages without any comments on the reason.

packages/sequent-core/src/services/pdf.rs
yuvalkom-M: i change it because this clippy error:
 error: this function's return value is unnecessarily wrapped by `Option`
   --> sequent-core/src/services/pdf.rs:620:9
    |
620 |         fn s3_bucket_path(path: String) -> Option<String> {
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   and also  this funciton is currently not in use.

packages/sequent-core/Cargo.toml
yuvalkom-M: this is the task requirement.

PR #2580, opened 2026-04-06
✨ Create rust '/metrics' endpoints for tracking business flows, to be used by prometheus

Findeton:
Parent issue: https://github.com/sequentech/meta/issues/7362

### DIFF
--- a/packages/Cargo.lock
+++ b/packages/Cargo.lock
@@ -4169,6 +4169,8 @@ dependencies = [
  "graphql_client",
  "handlebars",
  "immudb-rs",
+ "lazy_static 1.5.0",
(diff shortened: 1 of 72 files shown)


yuvalkom-M:
cargo clippy will throw errors on missing docs for every const like const METRIC_HTTP_REQUESTS_TOTAL: &str = "harvest_http_requests_total";
or struct like struct RequestStart(Instant);
and even for enum variant like Success in 
pub enum TaskOutcome {
    Success,
    Failure,
    Retry,
}

you can use #[allow(clippy::missing_docs_in_private_items)] for const and variant  if you like.

PR #2582, opened 2026-04-13

yuvalkom-M:
https://github.com/sequentech/meta/issues/11566


packages/harvest/src/routes/immudb_log_audit.rs
```diff
@@ -179,47 +215,47 @@ impl TryFrom<&Row> for PgAuditRow {
 
     fn try_from(row: &Row) -> Result<Self, Self::Error> {
         let mut id = 0;
-        let _audit_type = String::from("");
-        let mut class = String::from("");
-        let mut command = String::from("");
-        let mut dbname = String::from("");
+        let _audit_type = String::new();
```
eselimsen, line 218: Can we remove this var, can't find its usage?

packages/harvest/src/routes/insert_cast_vote.rs
```diff
@@ -45,36 +47,33 @@ pub async fn insert_cast_vote(
...
-            &format!("{:?}", e),
+            &format!("{e:?}"),
             ErrorCode::Unauthorized,
         )
     })?;
 
     info!("insert-cast-vote: starting");
 
-    let insert_result_wrapped = retry_with_exponential_backoff(
+    let insert_result_wrapped = Box::pin(retry_with_exponential_backoff(
```
eselimsen, line 57: Were these `Box::pin`s to resolve `large_futures` lint? If so, I think the above `#[allow(clippy::large_futures)]` becomes redundant?

packages/harvest/src/routes/users.rs
```diff
@@ -678,7 +735,7 @@ pub async fn edit_user(
     if voter_email_tlf_edit {
         /*check_edit_email_tlf(&client, &input, &realm, &new_attributes)
         .await
-        .map_err(|e| (Status::Unauthorized, format!("{:?}", e)))?;*/
+        .map_err(|e| (Status::Unauthorized, format!("{e:?}")))?;*/
```
eselimsen, line 738: Commented-out completely, shall we remove?

eselimsen (commented):
Judging only the changeset, and assuming arithmetical changes don't affect the business logic, I think this is a great starting point. Pointed out some points which I assume iteration artifacts.

PR #2587, opened 2026-04-20
✨ Keycloak: Configuration to support multiple certificates in Election Event

BelSequent:
Parent issue: https://github.com/sequentech/meta/issues/11110


docs/docusaurus/docs/01-system_introduction/01-system_introduction.md
xalsina-sequent, line 1: Please review this changes are not from another version.

xalsina-sequent (approved):

PR #2588, opened 2026-04-20

yuvalkom-M:
https://github.com/sequentech/meta/issues/11566

### DIFF
--- a/packages/step-cli/src/commands/cast_vote.rs
+++ b/packages/step-cli/src/commands/cast_vote.rs
@@ -2,13 +2,16 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
-use crate::{types::hasura_types::*, utils::read_config::read_config};
+use crate::types::hasura_types::uuid;
+use crate::utils::read_config::read_config;
 use clap::Args;
 use colored::Colorize;
 use graphql_client::{GraphQLQuery, Response};
+use tracing::{error, info};
 
 #[derive(Args)]
 #[command(about = "Cast a vote", long_about = None)]
+/// Cast ote command arguments
 pub struct CastVote {
     /// Election id - the election to cast a vote for
     #[arg(long)]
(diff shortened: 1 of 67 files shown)


packages/step-cli/src/commands/cast_vote.rs
```diff
@@ -2,13 +2,16 @@
...
+use crate::types::hasura_types::uuid;
+use crate::utils::read_config::read_config;
 use clap::Args;
 use colored::Colorize;
 use graphql_client::{GraphQLQuery, Response};
+use tracing::{error, info};
 
 #[derive(Args)]
 #[command(about = "Cast a vote", long_about = None)]
+/// Cast ote command arguments
```
eselimsen, line 14: Typo here?

packages/step-cli/src/commands/download_tally_results.rs
```diff
@@ -49,27 +53,27 @@ pub fn download_results(
...
     // Get the results event ID from the tally session execution
     let results_event_id = get_tally_session_execution::get_tally_session_execution(tally_id)?;
-    let results_event_id = match results_event_id {
-        Some(id) => id,
-        None => return Err(Box::from("No results event ID found for tally session")),
+
+    let Some(results_event_id) = results_event_id else {
+        return Err(Box::from("No results event ID found for tally session"));
     };
-    println!("Results event ID: {}", results_event_id);
```
eselimsen, line 56: Retain this log with `info!`, as it looks like part of the standard logging?

packages/step-cli/src/utils/tally/download_document.rs
```diff
@@ -78,13 +84,13 @@ pub fn download_file(
...
             .set_port(Some(port))
-            .map_err(|_| "failed setting port")?;
+            .map_err(|()| "failed setting port")?;
     }
 
     // Build the Host header exactly as it was originally signed
     let host_header = match original.port() {
-        Some(port) => format!("{}:{}", original.host_str().unwrap(), port),
-        None => original.host_str().unwrap().to_string(),
+        Some(port) => format!("{}:{}", original.host_str().unwrap_or_default(), port),
```
eselimsen, line 92: Is this correct as the expected behavior?

eselimsen (commented):
Judging only the changesets, and assuming arithmetical changes don't affect the business logic, I think this is a great starting point. Flagged a few points to ensure the intent. When avoiding panics, I think we should make sure to not fail/break silently.

PR #2594, opened 2026-04-23
✨ Voting Portal: Voter Facing Text Improvements

GuyZequent:
Parent Issue: https://github.com/sequentech/meta/issues/11924


packages/voting-portal/src/translations/en.ts
```diff
@@ -294,58 +281,67 @@ const englishTranslation = {
             },
         },
         errors: {
+            encoding: {
```
Findeton, line 301: as you're adding new fields here, you'll also need to add them to the other translations as en.ts defines the type.

packages/voting-portal/src/translations/en.ts
```diff
@@ -1,66 +1,53 @@
-// SPDX-FileCopyrightText: 2025 Sequent Tech Inc <legal@sequentech.io>
+// SPDX-FileCopyrightText: 2022 Félix Robles <felix@sequentech.io>
```
Findeton: The year should be 2026.

Findeton (changes requested):

packages/voting-portal/src/translations/en.ts
```diff
@@ -69,91 +56,91 @@ const englishTranslation = {
             descriptionNoAudit:
                 "To make changes in your selections, click “<b>Edit ballot</b>” button, to confirm your selections, click “<b>Cast your ballot</b>” button below.",
             backButton: "Edit ballot",
-            castBallotButton: "Cast your ballot",
+            castBallotButton: "Submit ballot",
             auditButton: "Audit ballot",
```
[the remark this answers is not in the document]
GuyZequent: Reverted terminology to "Cast"

packages/voting-portal/src/translations/en.ts
GuyZequent: Isn't this done automatically by i18? Do I need to manually translate and add this?

packages/voting-portal/src/translations/en.ts
```diff
@@ -21,55 +18,47 @@ const englishTranslation = {
...
-            warningDialog: {
-                title: "Review your ballot",
-                content:
-                    "Your ballot contains selections that may need your attention (such as selecting fewer options than allowed). Your ballot is valid and will be counted as submitted.",
-                ok: "Back and review",
+                    "Some of your answers will make one or more ballot questions invalid or blank",
+                ok: "Review selection",
                 continue: "Continue",
                 cancel: "Cancel",
             },
```
BelSequent, line 47: @GuyZequent `warningDialog` is removed from the file but not from the code, I guess this was not intended, shall I put it back?

BelSequent (changes requested):

packages/voting-portal/src/translations/en.ts
GuyZequent: Yes, that's a mistake on my part

BelSequent (approved):

PR #2613, opened 2026-05-05
✨Remove unsafe rust code

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/7111


eselimsen:
I have read the CLA Document and I hereby sign the CLA

packages/windmill/src/services/celery_app.rs
```diff
@@ -215,18 +226,17 @@ async fn create_connection() -> Result<(Arc<Connection>, String)> {
...
-        broker_connection_max_retries = BROKER_CONNECTION_MAX_RETRIES;
-        heartbeat = HEARTBEAT_SECS;
-    }
+    let CeleryConfig {
+        prefetch_count,
+        acks_late,
+        task_max_retries,
+        broker_connection_max_retries,
+        heartbeat_secs,
+        ..
```
BelSequent: I would remove the rest pattern (dots) here just to make it more robust. While it might seem convenient, in practice we would prefer to force the function to fail the compilation whenever we add a new parameter to the config, loki a kind of reminder "- Hey, there is a function here that deals with config stuff, you might want to do something with the new parameter"

packages/windmill/src/services/celery_app.rs
```diff
@@ -215,18 +226,17 @@ async fn create_connection() -> Result<(Arc<Connection>, String)> {
...
+    } = *CELERY_CONFIG
+        .read()
+        .map_err(|_| anyhow!("failed to read-lock CeleryConfig"))?;
```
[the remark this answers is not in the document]
BelSequent: Copilot wrongly assumes the compiler is creating a variable of type CeleryConfig which isn´t 
(there is no `let` expression for that variable). The destrcuturing is just copying the data members of CeleryConfig one by one because they all impl `Copy`

packages/windmill/src/services/celery_app.rs
```diff
@@ -93,74 +92,86 @@ impl Queue {
...
-pub fn get_queues() -> Vec<String> {
-    unsafe { QUEUES.clone() }
+/// Get global app execution status.
+pub fn get_is_app_active() -> bool {
+    *IS_APP_ACTIVE
+        .read()
+        .expect("failed to read-lock is_app_active")
 }
 
 lazy_static! {
```
BelSequent: Windmill is quiet slow to build and macros take a big piece of the cake.

Now that you are on this. We could remove the `lazy_static` dependency. `LazyLock ` is its modern replacement that's part of std and needs no extra crate.

I asked the AI to check, this is what She said:

_It's used in 8 places across the package, so it's not a one-liner removal — you'd need to migrate all of them. Each `lazy_static! { static ref FOO: T = ...; }` becomes `static FOO: LazyLock<T> = LazyLock::new(|| ...);.`

It's entirely doable since LazyLock is a drop-in replacement in all these cases, and then you could drop lazy_static from 

`services/import/import_tenant.rs:20:lazy_static! {
services/import/import_users.rs:31:lazy_static! {
services/import/import_scheduled_events.rs:23:lazy_static! {
services/export/export_users.rs:24:lazy_static! {
services/export/export_bulletin_boards.rs:27:lazy_static! {
services/database.rs:158:lazy_static! {
services/probe.rs:30:lazy_static! {
services/celery_app.rs:177:lazy_static! {
bin/main.rs:30:lazy_static! {`

But it's a refactor, not a one-file fix. Worth doing if you want to reduce dependencies, but no functional change._

BelSequent (changes requested):
Overall it looks very good. 
I just commented on minor inprovements that I think are worth to discuss. It can be quickly done: The rest pattern is one liner and the one about `lazy_static!` can be quickly refactored by the AI.

packages/windmill/src/services/celery_app.rs
eselimsen: I've did the refactor. Removed `async_once` and `lazy_static`, the former wasn't compatible with our use cases, so used `tokio::sync::OnceCell` instead. For testing, I ran a few background tasks from the admin portal (importing, for example), and ran all tests of `windmill` and `harvest`, with all features enabled, everything went fine.

BelSequent (approved):

PR #2626, opened 2026-05-06
✨ Constraint importing event into a different major version

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/11114


BelSequent (changes requested):
Now that you mentioned on the ticket the thing about the BACK_TRACE, I thought there is an opportunity to make this more user friendly:

Then endpoint already supports returning an error text + a message field, which will make more sense for a business logic like this instead of a raw error. If we do this way there is no backtrace on the UI.

`check_version_compatibility` does not need to return Anyhow (the backtrace), because the only invokation there is `extract_semver` which returns an Option, so a simple error with the text would serve and you can match this into the error field.

Unfortunately `check_version_compatibility` lives inside `get_election_event_schema`, which needs backtrace, but there is a quick workaround: I would just extract that encapsulation, and maybe add `map_err()` to the failure points with helpful messages.

BelSequent (approved):

BelSequent (approved):

PR #2634, opened 2026-05-07
🐞 Admin Portal > User attribute missing key

BelSequent:
Parent issue: https://github.com/sequentech/meta/issues/11994


xalsina-sequent (changes requested):
We have to use getTransaltionLabel in all the places where the attribute label is displayed:

In /user-roles for the admin users
- In the label for the columns displayed in the list
- In the label for in the select columns to display list
- in the label for the add filter list
- in the label for the filter
- in the label for the attribute when creating/edition user
Check also in the voter list for the election event:
- In the label for the columns displayed in the list
- In the label for in the select columns to display list
- in the label for the add filter list
- in the label for the filter
- in the label for the attribute when creating/editing a voter

Also apply the suggestion that copilot did as displayName could be null.

Image

xalsina-sequent (approved):

PR #2659, opened 2026-05-11
✨ IVR: Move to beyond

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/11109

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,128 +3,112 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 
 {
-	"name": "step",
-	"dockerComposeFile": "docker-compose.yml",
-	"service": "devcontainer",
-	"forwardPorts": [
-		"minio-proxy:9002",
-		"minio:9001",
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -3,128 +3,112 @@
...
+    "storage": "40gb",
+    "cpus": 4
+  },
+  "features": {
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {},
+    "ghcr.io/devcontainers/features/sshd:1": {
+      "version": "latest"
+    }
+  },
+  "customizations": {
```
Findeton, line 59: AFAIK the only real change is this so why the reordering of the rest of things?

```
"codespaces": {
  "repositories": {
    // Allow checking out private submodules
    "sequentech/beyond": {
      "contents": "read"
    }
  }
}
```

eselimsen: It was json reformatting/prettifying. I believe we won't use this anymore, I can revert this change, if you like.

PR #2680, opened 2026-05-18
🐞 Admin portal: stop click propagation missing when removing a candidate image

yuvalkom-M:
https://github.com/sequentech/meta/issues/11569


eselimsen (approved):
I recommend updating the title too.

PR #2683, opened 2026-05-18
🐞 Deleting election event does not remove the S3 event document files

BelSequent:
Parent issue: https://github.com/sequentech/meta/issues/11995


packages/sequent-core/src/services/s3.rs
```diff
@@ -1127,3 +1105,170 @@ mod tests {
         );
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{
```
yuvalkom-M, line 944: It seems that this file already contain tests and now they are duplicated

yuvalkom-M (changes requested):

yuvalkom-M (approved):

PR #2692, opened 2026-05-21
✨ IVR: Implement auth config fetching from Keycloak

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/12192


.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5-event-33f18502-a67c-4853-8333-a58630663559.json
```diff
@@ -763,6 +806,126 @@
...
+      "name": "IVR Service",
+      "description": "Used by the IVR to act on its own behalf, such as reading IVR auth configuration or blacklists.",
+      "rootUrl": "",
+      "adminUrl": "",
+      "baseUrl": "",
+      "surrogateAuthRequired": false,
+      "enabled": true,
+      "alwaysDisplayInConsole": false,
+      "clientAuthenticatorType": "client-secret",
+      "secret": "**********",
```
eselimsen, line 821: Exported secrets are masked like this, so not a problem.

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -0,0 +1,190 @@
...
+ * by an {@link AuthenticatorConfigModel} declaring the IVR metadata keys is a deployment-time
+ * misconfiguration and yields HTTP 500.
+ */
+@JBossLog
+public class IvrConfigResourceProvider implements RealmResourceProvider {
+    /**
+     * Voting client ID, it may have a specific flow override, in which case we prefer over realm's default flow.
+     * This specific ID is also used internally by the platform to auth voters through the IVR.
+     */
+    static final String IVR_VOTING_CLIENT_ID = "ivr-voting";
```
eselimsen: This is used around multiple times. I'd prefer a specific capability or similar, like the `ivr-config-read` below, but this is a well-known `client_id` that'll be used by the rest of the system as well (including conditional auth). We also have similar `azp`s like `voting-portal-kiosk` known specially by the system. I think it should be okay, at least for the initial phase.

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    private static AuthStep buildStep(RealmModel realm, AuthenticationExecutionModel exec) {
+        String authenticatorId = exec.getAuthenticator();
+        AuthStep stock = STOCK_AUTHENTICATORS.get(authenticatorId);
+        if (stock != null) {
+            return stock;
+        }
+
+        String configId = exec.getAuthenticatorConfig();
```
eselimsen: This is an optimistic, best-effort initial handling of such cases. Further testing and possible logic improvements would be in https://github.com/sequentech/meta/issues/12208.
Also, this doesn't authenticate anyone; only to collect the list of inputs we need from the client on the phone.

docs/docusaurus/docs/07-developers/12-ivr/09-ivr-config.md
```diff
@@ -0,0 +1,39 @@
+---
```
BelSequent, line 1: A few things I am thinking:
1. I don´t see how this can be a tutorial. The other tutorials describe a step by step fashion, but these clients are already configured by default so there is no need for it.
2. These are keycloak clients but there is no mention to keycloak in the doc.
3. The system design document is very long, perhaps good for now to feed AI but I would not rely on it to explain everything, I would keep adding short documents to explain components and eventually get rid or summarize the design doc.

I suggest to move this doc to `docs/docusaurus/docs/07-developers/06-keycloak` or `docs/docusaurus/docs/07-developers/12-ivr`, even better to split and have one brief document in each folder, the former for keycloak specific configurations and the later for IVR business logic (like to describe the components that are using these clients, link the other doc and so on).

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/AuthStep.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * <p>{@code promptKey} is optional and omitted from the JSON when null.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record AuthStep(
+    @JsonProperty("field") String field,
+    @JsonProperty("max_digits") int maxDigits,
+    @JsonProperty("terminator") String terminator,
+    @JsonProperty("maps_to") String mapsTo,
+    @JsonProperty("prompt_key") String promptKey) {}
```
BelSequent: We should have these props defined in one place within the extension.

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -0,0 +1,199 @@
...
+        .forEachOrdered(e -> steps.add(buildStep(realm, e)));
+
+    // If there are no auth steps "left" configured, it is very likely that was a configuration
+    // issue.
+    if (steps.isEmpty()) {
+      throw new WebApplicationException(
+          "There are no viable auth steps for IVR.", Response.Status.INTERNAL_SERVER_ERROR);
+    }
+
+    return Response.ok(Map.of("steps", steps)).build();
```
BelSequent: String literal not defined

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -0,0 +1,199 @@
...
+public class IvrConfigResourceProvider implements RealmResourceProvider {
+  /**
+   * Voting client ID, it may have a specific flow override, in which case we prefer over realm's
+   * default flow. This specific ID is also used internally by the platform to auth voters through
+   * the IVR.
+   */
+  static final String IVR_VOTING_CLIENT_ID = "ivr-voting";
+
+  /** Realm role required on the caller's token. */
+  static final String REQUIRED_ROLE = "ivr-config-read";
```
BelSequent, line 53: The design doc (§5.1.2, §C.8.b, checklist item 8) consistently says the required role is can_read_phone_blacklist — the same role that gates the blacklist Hasura read, intentionally shared so there is one role, one principal, one audit footprint for both pre-auth Lambda reads. But the implementation uses a different name.

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -0,0 +1,199 @@
...
+    RealmModel realm = session.getContext().getRealm();
+    AuthenticationFlowModel flow = effectiveDirectGrantFlow(realm);
+
+    List<AuthStep> steps = new ArrayList<>();
+    realm
+        .getAuthenticationExecutionsStream(flow.getId())
+        .filter(e -> !e.isAuthenticatorFlow()) // skip sub-flow references
+        .filter(IvrConfigResourceProvider::isRequiredOrConditional)
+        .filter(e -> !SKIPPED_AUTHENTICATORS.contains(e.getAuthenticator()))
+        .forEachOrdered(e -> steps.add(buildStep(realm, e)));
```
BelSequent, line 94: `.getAuthenticationExecutionsStream(flow.getId())` only returns direct children of the flow — not executions nested inside sub-flows. 

` !e.isAuthenticatorFlow()` filter then removes sub-flow container entries entirely. Any authenticators nested inside a sub-flow (e.g. OTP inside a CONDITIONAL sub-flow) are silently absent from the response. The Lambda gets an incomplete step list with no indication anything was omitted.

`IvrConfigResourceProvider::isRequiredOrConditional`. A CONDICTIONAL top-level execution is one that only runs when a condition evaluates to true at runtime. Including it unconditionally means the Lambda will always prompt the caller for that credential even when Keycloak won't require it for a given voter.

`SKIPPED_AUTHENTICATORS` is always empty.

BelSequent (changes requested):
Apart from my comments, there are also some new copilot notes worth checking

.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5-event-33f18502-a67c-4853-8333-a58630663559.json
```diff
@@ -763,6 +806,126 @@
...
+      ],
+      "webOrigins": [
+        "/*"
+      ],
+      "notBefore": 0,
+      "bearerOnly": false,
+      "consentRequired": false,
+      "standardFlowEnabled": false,
+      "implicitFlowEnabled": false,
+      "directAccessGrantsEnabled": false,
```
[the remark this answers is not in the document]
eselimsen: This client's implementation is part of https://github.com/sequentech/meta/issues/12174, but still, addressed with c5061338.

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -0,0 +1,199 @@
...
+    AccessToken token = extractToken();
+    if (token == null) {
+      log.debug("ivr-config: no bearer token");
+      throw new WebApplicationException(Response.Status.UNAUTHORIZED);
+    }
+
+    AccessToken.Access realmAccess = token.getRealmAccess();
+    if (realmAccess == null || !realmAccess.isUserInRole(REQUIRED_ROLE)) {
+      log.debug("ivr-config: token missing required realm role");
+      throw new WebApplicationException(Response.Status.FORBIDDEN);
```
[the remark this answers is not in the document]
eselimsen: I think checking the realm role is the correct approach, instead of tying to a specific client, so that's the intended behavior.

docs/docusaurus/docs/07-developers/12-ivr/09-ivr-config.md
eselimsen: Agreed, thanks for the pointers. Moved docs to `IVR` section, as this is an extension not part of the system (neither required for the base, so could be confusing if spread elsewhere). Also extended a bit to communicate the intent of it more, it was more of a "checklist" than a tutorial: 63d7009f

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
eselimsen: It was renamed for consistency with the codebase and our conventions, as requested by Felix too. The design doc is not complete, or fully accurate; the snippets are mostly illustrative examples, there are many changes over it.

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/AuthStep.java
eselimsen: Addressed in: 947e4a1a

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
eselimsen: Addressed in: 947e4a1a

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
eselimsen: Comments are correct; but the points are out of scope of this ticket, they will be addressed as part of https://github.com/sequentech/meta/issues/12208

eselimsen: Also mentioned here: https://github.com/sequentech/step/pull/2692#discussion_r3280939678

.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5-event-33f18502-a67c-4853-8333-a58630663559.json
```diff
@@ -763,6 +806,126 @@
...
+      "serviceAccountsEnabled": true,
```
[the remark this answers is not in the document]
eselimsen: Addressed in: c5b5e791

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -0,0 +1,199 @@
...
+   * default.
+   */
+  private static AuthenticationFlowModel effectiveDirectGrantFlow(RealmModel realm) {
+    ClientModel ivrClient = realm.getClientByClientId(IVR_VOTING_CLIENT_ID);
+    if (ivrClient != null) {
+      String overrideId = ivrClient.getAuthenticationFlowBindingOverride("direct_grant");
+      if (overrideId != null) {
+        AuthenticationFlowModel override = realm.getAuthenticationFlowById(overrideId);
+        if (override != null) {
+          return override;
```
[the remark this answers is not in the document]
eselimsen: Addressed in: 947e4a1a

BelSequent (approved):

PR #2696, opened 2026-05-25
✨ IVR: Add Hasura config for blacklisting

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/12170


hasura/metadata/rest_endpoints.yaml
```diff
@@ -1 +1,9 @@
-[]
+- comment: ""
+  definition:
+    query:
+      collection_name: allowed-queries
+      query_name: IvrIsPhoneBlacklisted
+  methods:
+    - GET
+  name: IvrIsPhoneBlacklisted
+  url: ivr/phone-blacklist/:event_id/:phone
```
eselimsen, line 9: Instead of using graphql directly on the IVR, added this simple endpoint. This allows us keep IVR leaner; less logic, less dependency, smaller size. Moving this business logic to here is also good, I think, from IVR's business perspective it is simply a lookup, we can change rules of it here if required, without change in IVR code or re-deployment.

.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5-event-33f18502-a67c-4853-8333-a58630663559.json
```diff
@@ -898,21 +978,162 @@
...
+        "display.on.consent.screen": "false",
+        "token.response.type.bearer.lower-case": "false",
         "dpop.bound.access.tokens": "false"
       },
       "authenticationFlowBindingOverrides": {},
       "fullScopeAllowed": true,
       "nodeReRegistrationTimeout": -1,
+      "protocolMappers": [
+        {
+          "id": "669ff7ce-ed0a-46f0-a065-c0fb40f1c432",
```
eselimsen, line 1011: Took `voting-portal` as reference.

hasura/migrations/backend-db/1779629886309_create_table_sequent_backend_phone_blacklist/up.sql
```diff
@@ -0,0 +1,21 @@
...
+        REFERENCES "sequent_backend"."tenant" ("id")
+        ON UPDATE CASCADE
+        ON DELETE CASCADE,
+    FOREIGN KEY ("tenant_id", "election_event_id")
+        REFERENCES "sequent_backend"."election_event" ("tenant_id", "id")
+        ON UPDATE RESTRICT
+        ON DELETE RESTRICT,
+    UNIQUE NULLS NOT DISTINCT ("tenant_id", "election_event_id", "phone_e164") -- Null election_event_id means tenant-wide, prevent duplicates
+);
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
```
BelSequent: Redundant CREATE EXTENSION pgcrypto

hasura/migrations/backend-db/1779629886309_create_table_sequent_backend_phone_blacklist/up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE "sequent_backend"."phone_blacklist"
+(
+    "id"                uuid        NOT NULL DEFAULT gen_random_uuid(),
+    "tenant_id"         uuid        NOT NULL,
+    "election_event_id" uuid,                                                  -- Null for tenant-wide blacklisting.
+    "phone_e164"        text        NOT NULL,
+    "reason"            text,
+    "created_at"        timestamptz NOT NULL DEFAULT now(),
+    "created_by"        uuid        NOT NULL,
```
BelSequent, line 9: **updated_at** column use to be in our tables, could we need it here?

hasura/metadata/databases/backend-db/tables/sequent_backend_phone_blacklist.yaml
```diff
@@ -0,0 +1,15 @@
...
+      permission:
+          columns:
+              - election_event_id
+              - phone_e164
+              - tenant_id
+          filter:
+              tenant_id:
+                  _eq: X-Hasura-Tenant-Id
+          allow_aggregations: true
+      role: phone-blacklist-read
```
BelSequent: Some details from claude I think worth to address, otherwise later on we will miss them.

- No admin-user role at all
- Every single other table (notification, preview, report, cast_vote, etc.) grants admin-user full CRUD or at minimum select
- No service-account role | Most tables give service-account unrestricted insert/delete for backend workers
- No write permissions for any role | There is no insert or delete permission defined — nobody can write to this table through Hasura
- The phone-blacklist-read select permission only exposes 3 columns (election_event_id, phone_e164, tenant_id), omitting id, reason, created_at, created_by. The IVR aggregate query only needs the count so it works, but it's unusual and would prevent any future query that needs reason or id from working under this role.

.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5.json
```diff
@@ -3306,58 +3336,58 @@
...
+            "introspection.token.claim": "true",
+            "userinfo.token.claim": "true",
+            "user.attribute": "trustee",
+            "id.token.claim": "true",
+            "lightweight.claim": "false",
+            "access.token.claim": "true",
+            "claim.name": "trustee",
+            "jsonType.label": "String"
+          }
+        },
```
BelSequent, line 3390: It is better to avoid indentation changes in git, just to have more clarity in case we need to look at it in the future.

BelSequent (changes requested):

hasura/migrations/backend-db/1779629886309_create_table_sequent_backend_phone_blacklist/up.sql
eselimsen: I meant it to be more of a append-only lookup table; either use it as created or recreate to change. I plan to make only the `reason` field changeable, which wouldn't require tracking, the rest would be read-only.

hasura/migrations/backend-db/1779629886309_create_table_sequent_backend_phone_blacklist/up.sql
eselimsen: I believe hasura adds this automatically. There were many duplicates in the past migrations (31 other), left it if it was a convention, will remove it happily if not. 👍 
86e57f0

hasura/metadata/databases/backend-db/tables/sequent_backend_phone_blacklist.yaml
eselimsen: 1. Hasura admin console interface works for manual changes, which is the intention for this stage. Admin portal interface is explicitly out of scope, it is part of this ticket: https://github.com/sequentech/meta/issues/12218
2. I think select constraints are correct. IVR doesn't need to know anything further, only lookup if a specific number has been blacklisted or not; no reason, id, audit fields or anything else is required, nor desired to be leaked to it.

.devcontainer/keycloak/import/tenant-90505c8a-23a9-4cdf-a26b-4e19f6a097d5.json
eselimsen: I agree, but the formatting of the file was broken, indents weren't consistent between objects, my auto-formatter fixed inconsistencies. I can revert these specific changes if you like (or make it a separate pr?), but I think better to have it fixed.

hasura/metadata/databases/backend-db/tables/sequent_backend_phone_blacklist.yaml
BelSequent: Better to leave the table's permissions ready, think this way... why would you want to revisit this again?.
- The phone-blacklist-read should be for the whole table, why would you have columns that will never be read by anyone?
- Hasura admin console is enough for now, but I would leave as much ready as I can otherwise the small details will accumulate. But If you are sure this is completely not a thing the admin-portal/backend will handle (unusual) then leave it as i is.
.

hasura/migrations/backend-db/1779629886309_create_table_sequent_backend_phone_blacklist/up.sql
BelSequent: I would add it, it does not harm and we donñt need another migration when we want to be able to update a blacklisted number.

hasura/metadata/databases/backend-db/tables/sequent_backend_phone_blacklist.yaml
eselimsen: Per our discussion, added the permissions for `admin-user` and `service-account` per conventions: 381f55f7

hasura/migrations/backend-db/1779629886309_create_table_sequent_backend_phone_blacklist/up.sql
eselimsen: Added conventional `updated_at` column per our discussion: 25f4e64 deef8cd

BelSequent (approved):

PR #2697, opened 2026-05-26
✨ Consistent Invalid Vote Policy

yuvalkom-M:
https://github.com/sequentech/meta/issues/12124


xalsina-sequent (changes requested):
Ballot-Verifier
- ballot-verifier Shows declined to vote selections as blank vote.

Sequent Core
- Some AI comments to be removed.
- The bit for decline to vote is added in encoding regardless of policy. Could be bad for backwards compatibility.
- Some parenthesis in a if condition.
- Gaps in unit tests
    * Test for test_multi_contest_reencoding_with_invalid_decline_to_vote does not check 
    * No test verifies that is_decline_to_vote is correctly mapped from the ballot-level flag in map_decoded_ballot_choices_to_decoded_contests (plaintext.rs). The mapping is a semantically important inversion — decoded_ballot_choices.is_explicit_invalid → is_decline_to_vote — with no unit test.
    * No test for get_bases() output shape (length and base values).
    * No test for the choices vector layout (verifying the per-contest flag sits at the expected index for each contest).
    * random_choice_for_contest in tests always sets is_explicit_invalid: false. test_roundtrip and test_mixed_radix_encode never exercise is_explicit_invalid: true in the random path

Admin portal Tally
- Put the participation percentage as the rightmost column.

Reports/Tally
- A voter who intentionally declined to vote is being counted as an under-vote.

Documentation
- Add documentation about how tu use decline to vote policy and how it affects admin-portal, voting-portal, results.

xalsina-sequent (approved):

PR #2710, opened 2026-05-28
🐞 Windmill - Voter import fails with unique constraint error when keycloak realm has duplicate emails enabled

xalsina-sequent:
Parent issue: https://github.com/sequentech/meta/issues/12177

### DIFF
--- a/packages/windmill/src/postgres/keycloak_realm.rs
+++ b/packages/windmill/src/postgres/keycloak_realm.rs
@@ -55,6 +55,13 @@ pub async fn get_realm_id(
         .clone())
 }
 
+/// Returns whether duplicate email addresses are permitted within the given Keycloak realm.
+///
+/// Queries the `duplicate_emails_allowed` flag directly from the Keycloak `realm` table.
+/// 
+/// # Errors
+/// 
+/// Returns an error if the realm is not found or the column cannot be read.
 #[instrument(skip(keycloak_transaction), err)]
 pub async fn get_duplicate_emails_allowed(
     keycloak_transaction: &Transaction<'_>,
--- a/packages/windmill/src/postgres/keycloak_realm.rs
+++ b/packages/windmill/src/postgres/keycloak_realm.rs
@@ -58,9 +58,9 @@ pub async fn get_realm_id(
 /// Returns whether duplicate email addresses are permitted within the given Keycloak realm.
 ///
 /// Queries the `duplicate_emails_allowed` flag directly from the Keycloak `realm` table.
-/// 
+///
 /// # Errors
-/// 
+///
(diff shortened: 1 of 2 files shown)


yuvalkom-M (commented):
there is a copilot comment you might want to address
also the new function need a documentation (comment above)

yuvalkom-M (approved):

packages/windmill/src/postgres/keycloak_realm.rs
```diff
@@ -54,3 +54,26 @@ pub async fn get_realm_id(
...
         .clone())
 }
+
+/// Returns whether duplicate email addresses are permitted within the given Keycloak realm.
+///
+/// Queries the `duplicate_emails_allowed` flag directly from the Keycloak `realm` table.
+///
+/// # Errors
+///
+/// Returns an error if the realm is not found or the column cannot be read.
```
BelSequent, line 64: Prefer `<br>` instead of new lines.
Also add it to Claude.md

packages/windmill/src/postgres/keycloak_realm.rs
```diff
@@ -54,3 +54,26 @@ pub async fn get_realm_id(
...
+    let row = keycloak_transaction
+        .query_one(
+            "SELECT duplicate_emails_allowed FROM realm WHERE id = $1",
+            &[&realm_id],
+        )
+        .await
+        .context("Error querying duplicate_emails_allowed")?;
+    row.try_get::<_, bool>("duplicate_emails_allowed")
+        .context("Error reading duplicate_emails_allowed")
+}
```
BelSequent, line 79: I think we should stop growing this pattern that does the directly accessing to postgres-keycloak. Check the effort analysis to do this using the KC API.

[duplicate-emails-allowed-api-analysis.md](https://github.com/user-attachments/files/28749275/duplicate-emails-allowed-api-analysis.md)

BelSequent: Yeah and also worth to add this rule into Claude.md

packages/windmill/src/services/import/import_users.rs
```diff
@@ -486,7 +490,7 @@ pub async fn import_users_file(
     let keycloak_transaction = keycloak_db_client
         .transaction()
         .await
-        .map_err(|err| anyhow!("Error starting Keycloak transaction: {err}"))?;
+        .context("Error starting Keycloak transaction")?;
```
BelSequent, line 493: This by itself will not work becasue there are 2 `map_err` above in the call hierarchy doing a format like `{err}` which take the context message but not the details and stacktrace. For that we need to format like `{err:?}` on those as well.

BelSequent (changes requested):

PR #2723, opened 2026-06-01
✨ Browser Based Trustees: Symmetric keys for each event in the browser-based trustee

BelSequent:
Parent issue: https://github.com/sequentech/meta/issues/10354


docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
```diff
@@ -0,0 +1,604 @@
...
+---
+
+## Service Communication Paths
+
+Braid-wasm is considered part of admin-portal.  Rows marked **BBT** are new or changed by
+this proposal; all others are unchanged from the current architecture.
+
+| # | Origin | Destination | Protocol | What |
+|---|--------|-------------|----------|------|
+| 1 | Admin-portal | Harvest | HTTP POST | `create-keys-ceremony`, `list-keys-ceremonies`, `get-private-key`, `check-private-key`, all admin operations |
```
Findeton: Yes but it's a post to hasura that in the end reaches harvest.

docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
```diff
@@ -0,0 +1,604 @@
...
+
+`get_trustee_mode_policy()` in `windmill/src/postgres/trustee.rs` already reads the
+`TrusteeModePolicy` from the trustee's `annotations.trustee_mode_policy` JSON field, so the
+mode is already detectable.  The query in the document's Component Changes section
+(`ORDER BY election_event_id NULLS LAST`) covers both cases in one SQL statement, but the
+calling code must know whether to pass an `election_event_id` parameter at all.
+
+### Options
+
+> **Recommended: Option C — see below.**
```
Findeton: Just keep option C

docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
```diff
@@ -0,0 +1,604 @@
...
+`create_keys_ceremony` creates the record normally (`STARTED`).  `create_keys_impl` checks
+whether every BBT trustee in the ceremony has a `public_key`; if not, it returns `Ok(())`
+without advancing the status.  On the next beat cycle `process_board` dispatches `create_keys`
+again and the check is retried automatically.
+
+Pro: no new status needed; ceremony advances on its own once trustees are ready.
+Con: ceremony appears stuck in `STARTED` with no UI feedback; admin cannot distinguish
+"waiting for trustees" from a genuine failure.
+
+**Option C — Rename `USER_CONFIGURATION` to `AWAITING_TRUSTEE_KEYS` ✓ Recommended**
```
Findeton: We need to be clear about the changes. Currently, posting the config message to the bulletin board happens synchronously, harvest performs the job in the call that creates the Keys Ceremony. In this new case, FOR BBT case only, the WINDMILL task that advances the Keys Ceremony will have the job of advancing the Keys Ceremony to "Started" AND at the same time posting the config message to the bb when all browser trustees have created and uploaded their keys.

docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
```diff
@@ -0,0 +1,604 @@
...
+`KeysCeremonyExecutionStatus::CANCELLED` is defined in the enum in `sequent-core` but is
+**dead code**: no Harvest endpoint, no Windmill task, and no service function ever writes it.
+Keys ceremony cancellation does not exist today.
+
+A future implementation would require at minimum:
+
+- A new Harvest endpoint (e.g. `POST /cancel-keys-ceremony`) that validates the current
+  `execution_status` is not already `CANCELLED` and that the election has not been started yet, then writes `CANCELLED`.
+  It should be allowed to cancel on `SUCCESS` but only if the voting period has not been started.
+- A decision on what to do with the B4 board if DKG was already `IN_PROGRESS`: the board
```
Findeton: We just leave the messages orphaned. It doesn't matter because the ceremony only advances when the browser trustees work on them, and for that they need to input the keys from the admin portal and they won't be able to do that when the ceremony is cancelled.

docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
```diff
@@ -0,0 +1,604 @@
...
+  already holds `Channel`, `Shares`, and possibly `PublicKey` messages.  These cannot be
+  deleted from an append-only board, so a cancelled ceremony would leave orphaned board
+  messages.  The admin would need to create a fresh ceremony on a new board.
+- For BBTs: a decision on whether to delete the `(trustee_id, election_event_id)` rows from
```
Findeton: We just leave the rows. Nevertheless a new keys ceremony will mean the trustees public keys will be different/regenerated. The trustee keys are not just (trustee_id, election_event_id), they should be (trustee_id, election_event_id, keys_ceremony_id)

docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
```diff
@@ -0,0 +1,604 @@
...
+    signing_key_pk:  string,   // StrandSignaturePk::from_sk()  → base64-DER Ed25519 public key
+    encryption_key:  string,   // symm::gen_key()               → base64 AES-256 symmetric key
+  }
+```
+
+This produces the same three values found in `trustee1.toml`.  All underlying functions
+(`StrandSignatureSk::generate`, `symm::gen_key`) are already present in braid and used in
+`gen_trustee_config.rs` — no new crypto dependency.
+
+### 2 — DB schema: add `election_event_id` to `sequent_backend_trustee`
```
Findeton: it should add `keys_ceremony_id` as well and both values, although nullable, should be Foreign Keys to the ids of the respective tables.

docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
```diff
@@ -0,0 +1,604 @@
...
+
+### Key-reading must be mode-aware
+
+Once the `election_event_id` column is added to `sequent_backend_trustee`, the logic that
+reads public keys in `create_keys_impl` must differentiate by trustee mode:
+
+| Trustee type | How to read `public_key` |
+|---|---|
+| **Server-based** | Read the global `public_key` directly — no `election_event_id` filter. |
+| **Browser-based** | Filter by both `tenant_id` **and** `election_event_id` to get the per-event key. |
```
Findeton: And keys_ceremony_id

docs/docusaurus/docs/07-developers/09-braid/key_ceremony_bbt_propossal_no_protoccol_change.md
BelSequent: Not exactly, the create_keys_ceremony does not publish the config message.
I have updated the doc to show more details.

Findeton (changes requested):
I think this proposal needs a few updates to reflect the intended BBT ceremony flow.

1. **Section: Flow / Component Changes / Scope note — ceremony-scoped trustee work.**  
   Browser trustee work should be scoped to a specific ceremony, not just to opening the election event page. The Flow currently says “BBT trustee opens election event in admin portal” and then `HeadlessTrusteeProvider` mounts (lines 370-371), and Component Changes says keys are generated/register on mount using `bbt_keys_{electionEventId}` (lines 576-583). The intended behavior is that trustee browser work starts when the trustee clicks into a specific keys ceremony. This is also consistent with tally: trustees need to input/restore their keys for the tally ceremony too, so both keys and tally ceremonies should only advance when the trustee user is inside that specific ceremony and actively participating.

2. **Section: Flow — download timing.**  
   The Flow puts the Download step after Windmill posts `Configuration`, browsers run DKG, and all trustees reach `KEY_GENERATED` (lines 386-402). That matches the intended rule that trustees can download the TOML/identity-key backup only once all browser trustees have generated their keys. However, the doc uses “keys” for both identity keys and DKG-generated keys, so it should explicitly say: download is enabled only after all selected trustees have reached `KEY_GENERATED`.

3. **Section: Ceremony Creation Gate / Flow — state transition inconsistency.**  
   “How it works with this change” says once all keys are present, Windmill posts `Configuration` and moves directly `AWAITING_TRUSTEE_KEYS -> IN_PROGRESS` (lines 148-151, 167-177). But the Flow still says `AWAITING_TRUSTEE_KEYS -> STARTED`, then `create_keys` posts `Configuration`, then `IN_PROGRESS` (lines 381-389). One of these should be corrected.

4. **Section: Cancellation Window — cancellation until related election(s) open.**  
   The doc says keys ceremony cancellation is “not currently implemented” (lines 313-317), then proposes cancellation if the election/voting period has not started (lines 321-323). That matches the intended rule that a ceremony is cancellable until the related election(s) are open, but it should say “related election(s)” explicitly and define the multi-election condition. Also the transition table says `SUCCESS` is terminal with no cancellation (lines 336-342), which conflicts with line 323 saying cancellation on `SUCCESS` should be allowed before voting starts.

5. **Section: Cancellation Window / Key Loss — unable to download or lost TOML.**  
   The doc should explicitly state that if a trustee cannot download the TOML/identity backup, or loses it and cannot restore it, the ceremony can be cancelled and recreated as long as the related election(s) are not open. Today this is only implied by the orphan-board-message discussion (lines 324-329) and the restore-from-backup discussion (lines 519-525).

6. **Section: Flow / What the verify step triggers — reusable key check.**  
   The Check step currently appears only as part of the linear Download/Check ceremony flow (lines 404-417). The intended behavior is broader: after keys have been generated, trustees should always have a way to upload/check their TOML/identity backup for validity. The doc should clarify whether this validation flow is available after `KEY_GENERATED`, after `KEY_CHECKED`, and during later tally ceremony preparation.

Also, I'd make the sections numbered, otherwise it's difficult to talk about them.

Findeton (commented):
I think the BBT backup/verify flow needs to be specified more concretely, especially after the security discussion with Eduardo.

For security, the browser should **never store the TOML / trustee secret material in `localStorage` or `sessionStorage`**. The TOML contents should exist only in memory while the trustee is inside the ceremony page. If the trustee tries to close or navigate away before confirming the download/check step, the browser should show a warning that leaving may make the ceremony unrecoverable for that trustee.

For the verify flow, Harvest must not receive the TOML because it contains secret material. Instead, we should use a browser-local verification plus signed attestation:

1. Harvest exposes a Hasura/GraphQL action to issue a short-lived verification nonce.
2. The browser parses the uploaded TOML locally.
3. The browser verifies locally that `signing_key_sk` derives the expected `signing_key_pk`.
4. The browser computes an `encryption_key_commitment`, e.g. an HMAC/hash commitment over ceremony-scoped data using the TOML `encryption_key`.
5. The browser signs a canonical verification payload with `signing_key_sk`, including:
   - `tenant_id`
   - `election_event_id`
   - `keys_ceremony_id`
   - `trustee_id`
   - `signing_key_pk`
   - `encryption_key_commitment`
   - Harvest-issued nonce
6. The browser sends only the signed payload, signature, nonce, and commitment to Harvest via Hasura/GraphQL. It never sends `signing_key_sk`, `encryption_key`, or the TOML file.
7. Harvest verifies:
   - the JWT trustee matches the ceremony trustee
   - the nonce was issued, is unexpired, and unused
   - the signature verifies under the registered `signing_key_pk`
   - the submitted commitment matches the commitment previously recorded for this trustee/ceremony
8. If valid, Harvest records `KEY_CHECKED`.

This also means the commitment needs to be recorded in Hasura earlier, when the trustee confirms the TOML download / key backup. The nonce should also be recorded in Hasura, but only as temporary challenge state with expiry/used status.

On B4 posting: the browser trustee should still sign B4 messages locally. Harvest must never sign on behalf of the BBT trustee. If we want Harvest in the path for audit/status reasons, it can only relay an already-signed B4 message; otherwise the browser can keep posting directly to B4 and separately notify Harvest/Hasura for ceremony state and commitment recording.

Finally, because the browser will not persist the TOML in web storage, cancellation/retry UX becomes important. A keys ceremony should be easily cancellable while the related election(s) are not open, and there should be a duplicate/recreate option when the related elections do not currently have a keys ceremony assigned. That assignment is represented by the keys ceremony reference on the election row, so cancellation/duplication needs to account for whether those election rows already point to a ceremony.

Findeton (changes requested):
Hi @BelSequent . Second pass on the doc after talking with David:
1. Drop the nonce/commitment/attestation apparatus entirely, including all "prove/proof" language. We can't prove possession of a symmetric key, and the commitment only ever verified a self-report. Check Step becomes browser-local validation plus a plain JWT-authenticated status advance; say plainly that KEY_RETRIEVED/KEY_CHECKED are trustee-confirmed, not proven. You know the sections; nice side effect is the migration shrinks to the two FK columns.
2. New section: show trustee public keys in the ceremony UI. This will require a small ui design from Figma. Info (ℹ) icon per trustee row in the ceremony trustee tables → popup showing the full public-key text block, for every trustee (server-based included, the point is verifying the complete set that goes into the Configuration). Trustee users verify with each other, out of band, that they all see identical keys. One line of justification max: the software is open source and the trustee trusts the software they run. Two refinements worth writing down: once the Configuration is on the board, display the key from the board rather than the DB row (the board is the frozen, auditable artifact, and sessions read it anyway), and state when comparison is meaningful - from IN_PROGRESS onward - naturally at the Download/Check stage. Needs a small Figma design (icon + popup); flag it as a dependency.
3. Unify the registration path now (replaces the "accepted divergence" framing). Make "per-ceremony registered key" the single mechanism for both trustee types: BBT registers interactively; for server-based trustees, windmill writes their stable key into a per-ceremony row at ceremony creation. Result: one gate ("every trustee has a key registered for this ceremony", server rows are just instant), one read query with no NULL-fallback, uniform cancellation (stale rows scoped by keys_ceremony_id for everyone), and snapshots protect tally from any future server-key rotation. A closer-to-final variant - braid-native self-registering through the same Harvest endpoint using its authorized-boards claim - exists but touches braid-native; defer it to the catch-up with David. Then keep a short, honest trade-offs note on what still can't be unified without protocol change: the key registry is mutable DB rather than the append-only board, windmill builds the Configuration from DB, and the trustee↔key binding rests entirely on the Keycloak session. That's the real remaining debt.
4. Convergence section pointing at David's branch. His follow-up (which also changes the encryption itself, so it can't merge yet) adds a protocol-level registration message: each trustee generates their toml-equivalent and posts it to the board, and the Configuration is built from those messages; server trustees follow the exact same path, their message just always has identical content. With point 3 in place the migration is mechanical, the per-ceremony DB row is the stand-in for the on-board registration message, and the windmill auto-snapshot becomes braid-native posting its constant registration. Say this explicitly so the design stays migration-shaped, and add a pointer in the protocol-change doc that David's branch supersedes/extends it.
5. Short "future work: certificate chain" note. Per-ceremony trustee certs will eventually chain upward (ie trustee → tenant → environment → Sequent root ) which is what gives fresh per-ceremony keys verifiable meaning; the on-screen comparison is the stopgap until then.

PR #2725, opened 2026-06-02
 🐞 Windmill - Windmill silently stops while importing a big event with 150 elections

yuvalkom-M:
https://github.com/sequentech/meta/issues/12245


packages/windmill/src/postgres/candidate.rs
```diff
@@ -50,45 +74,62 @@ pub async fn insert_candidates(
...
-            "#,
-        )
-        .await?;
+        let id = parse_uuid_v4(&candidate.id)?;
+        let contest_id = candidate
+            .contest_id
+            .as_ref()
+            .and_then(|contest_id| parse_uuid_v4(contest_id).ok());
+
+        let row: Vec<&(dyn ToSql + Sync)> = vec![
```
eselimsen: Because this is a performance improvement PR, I think worth to mention: this will allocate a new vec for each `candidate`, right? I think, it would be better to simply use an array:
```rust
let row: [
    &id as &(dyn ToSql + Sync),
    &tenant_uuid,
    <...the-rest...>
]
```

eselimsen (changes requested):

eselimsen (approved):

PR #2726, opened 2026-06-02
✨ IVR: Implement voter auth and eligibility check

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/12174


packages/admin-portal/src/resources/Settings/SettingsVotingChannel.tsx
```diff
@@ -37,6 +37,7 @@ export const SettingsVotingChannels: React.FC<void> = () => {
     const [voting, setVoting] = useState<any>({
         online: record?.voting_channels?.online || true,
         kiosk: record?.voting_channels?.kiosk || false,
+        telephone: record?.voting_channels?.false || false,
```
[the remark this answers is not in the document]
eselimsen: Was fixed with 10aab9d0

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/AuthStep.java
```diff
@@ -17,6 +17,6 @@
 public record AuthStep(
     @JsonProperty(Constants.AUTH_STEP_PROP_FIELD) String field,
     @JsonProperty(Constants.AUTH_STEP_PROP_MAX_DIGITS) int maxDigits,
-    @JsonProperty(Constants.AUTH_STEP_PROP_TERMINATOR) String terminator,
+    @JsonProperty(Constants.AUTH_STEP_PROP_KIND) String kind,
     @JsonProperty(Constants.AUTH_STEP_PROP_MAPS_TO) String mapsTo,
```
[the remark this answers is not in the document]
eselimsen: It is not deployed yet, not a problem.

packages/keycloak-extensions/ivr-config-provider/src/main/java/sequent/keycloak/ivr_config/IvrConfigResourceProvider.java
```diff
@@ -176,11 +178,15 @@ private static AuthStep buildStep(RealmModel realm, AuthenticationExecutionModel
     }
     String fieldName = c.get(Constants.AUTH_STEP_PROP_FIELD);
     String mapsTo = c.get(Constants.AUTH_STEP_PROP_MAPS_TO);
-    if (fieldName == null || mapsTo == null) {
-      String msg = "AuthenticatorConfig for '%s' is missing required IVR keys (%s, %s)";
+    String kind = c.get(Constants.AUTH_STEP_PROP_KIND);
+    if (fieldName == null || mapsTo == null || kind == null) {
+      String msg = "AuthenticatorConfig for '%s' is missing required IVR keys (%s, %s, %s)";
```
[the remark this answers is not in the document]
eselimsen: It is not deployed yet, not a problem.

eselimsen (commented):
I've found some design gaps and improvement opportunities in the design doc; for the background and agreements, please make sure to read the ticket: https://github.com/sequentech/meta/issues/12174

hasura/metadata/query_collections.yaml
```diff
@@ -10,3 +10,13 @@
                     }
                   }
                 }
+          - name: IvrGetElectionStatus
+            query: |
+                query IvrGetElectionStatus($election_id: uuid!) {
```
eselimsen, line 15: `sequent_backend_election`'s pk definition contains `tenant_id` and `election_event_id` too, so we do direct query here. Because hasura row-level permissions will filter by `tenant_id`, there shouldn't be a risk of leak.

(I'm not sure if `election`s current pk is correct, I assume uu`id` should be unique, it was changed to composite in the following commit: https://github.com/sequentech/step/commit/b235e6296eeeca22925e44cfce01d6b72397e47a, please correct me if I'm wrong, I can add `election_event_id` parameter if required)

BelSequent: Good point, you are right the uuids are unique.
However there are 2 separate things, the first you mentioned about the pk correctness and the 2nd thing about adding more parameters to the filter.

The 1st is out of scope here but It´s done on purpose, so like a structural decision... that way it forces the foreign keys to know which tenant they are querying and so on.

The 2nd thing: I would use parameters (tenant_id, election_event_id, id), although uuids are unique but this way we gain more defence and index performance, I guess that´s the reason we do it in nearly every table in windmill postgres.

packages/sequent-core/src/ballot.rs
BelSequent, line 1883: We should define how the new `VotingStatusChannel::TELEPHONE` will fit into the current business logic.
Actually I think we need a better definition of what a **channel** is and what is not.
Otherwise this confussion will compound, I mean by that: 

Currently the admin function "Start Voting" can be triggered for any channel independently, but there are some interdependency rules...
Once `ONLINE` voting is started the `EARLY_VOTING` is stopped. One could expect the `TELEPHONE `voting to also stop  `EARLY_VOTING`.

Another open question (which is the same question but in the other direction - we need a better definition) would be: 
Shall the `TELEPHONE_VOTING` be allowed during `EARLY_VOTING` ?

Anyway, I think this might be out of scope for this ticket so I will create another one.

BelSequent (changes requested):

hasura/metadata/query_collections.yaml
eselimsen: Copying our discussion for future reference:

The question was about pk for "uniqueness" assumption: identifying the row reliably without duplicates, to confirm that there are no business cases that would cause duplicated values for the `id` field alone. This is confirmed, and it is safe to do `id`-only lookups as here.

While we also agree on composite pk for db integrity, `id` is assumed, and should be unique too, we need a unique constraint to enforce this. For index performance, I believe reordering the composite pk order would help for `tenant_id` + `election_event_id` based lookups due to leftmost rule on pg btree indexes.
I've created a ticket to address these: https://github.com/sequentech/meta/issues/12294

packages/sequent-core/Cargo.toml
```diff
@@ -151,6 +151,9 @@ plugins_wit = ["dep:wasmtime", "dep:wasmtime-wasi"]
 default_features = ["dep:strand", "dep:num-bigint", "dep:tempfile", "dep:ammonia"]
 
 sqlite = ["dep:tokio", "dep:rusqlite"]
+time = ["dep:time"]
+jwt = ["time"]
+tracing = ["dep:tracing"]
```
BelSequent, line 158: This is sound. The only thing I see worth improving is that the `jwt ` feature could contain time and tracing, so it´s complete by itself.
Tracing is always available whenever jwt is compiled, because keycloak activates both, that's why it builds but is weak.
You made it better anyway.

BelSequent (changes requested):

packages/sequent-core/Cargo.toml
eselimsen: Done in 2cb67c1
Confirmed compilation by building `harvest`, `windmill` and `velvet` locally, and rebuilding devcontainer from scratch (keeping the volumes though).

PR #2731, opened 2026-06-02
🐞 Admin Portal: Double Downloads

yuvalkom-M:
https://github.com/sequentech/meta/issues/12248


yuvalkom-M:
After re-review: the bug doesn't happen in 9.0

PR #2819, opened 2026-06-30
✨ Admin Portal: Configurable Date/Time Format for Voting Portal

BelSequent:
Parent issue: https://github.com/sequentech/meta/issues/12355


.github/workflows/tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           - service: immu-board
           - service: immudb-rs
           - service: sequent-core
-            extra: --features keycloak
+            extra: --features keycloak,default_features
```
BelSequent, line 26: Without this we are not covering all the tests in GA

packages/ui-core/tsconfig.json
```diff
@@ -29,5 +29,11 @@
   },
   "include": [
     "src"
+  ],
+  "exclude": [
+    "node_modules",
+    "dist",
+    "src/**/*.test.ts",
+    "src/**/*.test.tsx"
```
BelSequent, line 37: Exclude test/build files from the production build, also to avoid rebuid issues.

packages/admin-portal/src/resources/ElectionEvent/EditElectionEventDataForm.tsx
```diff
@@ -766,6 +847,7 @@ export const EditElectionEventDataForm: React.FC = () => {
                 defaultValues={defaultValues}
                 validate={formValidator}
                 record={parsedValue}
+                resetOptions={{keepDirtyValues: true, keepErrors: true}}
```
xalsina-sequent: If we set KeepDirtyValues: true causes the form to think that the values have changed even after a save. The first time you click save after changing a value the save button is not disabled allowing you to click it again which causes errors as no change is available to save.

We should remove resetOptions.

xalsina-sequent (changes requested):

xalsina-sequent (approved):

PR #2823, opened 2026-07-02
 ✨ IVR: Update election event creation automation for IVR

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/12214


packages/windmill/src/services/import/import_election_event.rs
```diff
@@ -228,17 +230,39 @@ pub fn read_default_election_event_realm() -> Result<RealmRepresentation> {
...
+    let keycloak_client_id = env::var("KEYCLOAK_CLIENT_ID")
+        .ok()
+        .map(|s| s.trim().to_string())
+        .filter(|s| !s.is_empty())
+        .with_context(|| "KEYCLOAK_CLIENT_ID can't be empty")?;
+    let keycloak_client_secret = env::var("KEYCLOAK_CLIENT_SECRET")
+        .ok()
+        .map(|s| s.trim().to_string())
+        .filter(|s| !s.is_empty())
+        .with_context(|| "KEYCLOAK_CLIENT_SECRET can't be empty")?;
```
eselimsen, line 245: This adds a new failure mode, but I think this is the correct behavior, we shouldn't accept empty-ish strings as secret values anyway.

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+[toolchain]
+channel = "1.96.0"
```
eselimsen, line 5: @xalsina-sequent Sneaked this in here, as mentioned recently.

PR #2850, opened 2026-07-09
✨ IVR: Admin interface for overriding system prompts

eselimsen:
Parent issue: https://github.com/sequentech/meta/issues/12412


packages/admin-portal/src/resources/ElectionEvent/IvrPrompts.tsx
```diff
@@ -0,0 +1,553 @@
...
+import Add from "@mui/icons-material/Add"
+import {ResourceListStyles} from "@/components/styles/ResourceListStyles"
+import {cloneDeep} from "lodash"
+import {ElectionHeaderStyles} from "@/components/styles/ElectionHeaderStyles"
+
+const RESOURCE = "sequent_backend_election_event"
+
+type Prompts = Record<string, Record<string, string>>
+
+const collectRequiredPromptKeys = (configAnnotation: string) => {
```
Findeton, line 50: What's the returning format of this?

packages/admin-portal/src/resources/ElectionEvent/IvrPrompts.tsx
```diff
@@ -0,0 +1,553 @@
...
+                    setDeleteId(id)
+                    setOpenDelete(true)
+                }
+            },
+            showAction: (id) => typeof id !== "string" || !requiredPromptKeys.has(id),
+        },
+    ]
+
+    // Editor operations
+    const createPromptKey = (rawKey: any, rawValue: any) => {
```
Findeton, line 272: returning type?

packages/admin-portal/src/resources/ElectionEvent/IvrPrompts.tsx
```diff
@@ -0,0 +1,553 @@
...
+                    newData[lang] = {}
+                }
+                if (!(key in newData[lang])) {
+                    newData[lang][key] = value
+                }
+            })
+            setEditorData(newData)
+        }
+    }
+    const updatePromptKey = (rawValue: any) => {
```
Findeton, line 289: Why any?

eselimsen (commented):
For future reference: `any`s were used because that's what we actually get from the external component we use, and the they checks were implemented inside the respective functions, so that's the correct type we really get and the reason I used it.
I'll create a follow-up ticket to mark them with the expected shapes instead, as you requested.

PR #2866, opened 2026-07-15
✨ Publish Election Results to the voting portal

Findeton:
Parent issue: https://github.com/sequentech/meta/issues/472


packages/results-portal/src/hooks/useAuthenticatedResults.ts
```diff
@@ -0,0 +1,209 @@
...
+                    loading: false,
+                    error: "Authentication failed",
+                }
+            }
+
+            await session.keycloak.updateToken(30).catch(() => undefined)
+
+            return authStateFromSession(session)
+        } catch (error) {
+            session.initPromise = undefined
```
xalsina-sequent: Clearing only initPromise leaves the failed Keycloak instance cached, but keycloak-js does not support calling init() twice on one instance. Please evict this session so a retry creates a fresh client.

packages/admin-portal/src/resources/ElectionEvent/EditElectionEventDataForm.tsx
```diff
@@ -1232,6 +1266,41 @@ export const EditElectionEventDataForm: React.FC = () => {
...
+                            source={"resultsWebsitePolicy.access"}
+                            choices={resultsWebsiteAccessOptions()}
+                            label={t("tally.resultsPublication.policyAccess")}
+                            defaultValue={EResultsWebsiteAccess.PUBLIC}
+                            emptyText={undefined}
+                            validate={required()}
+                        />
+                        <SelectInput
+                            source={"resultsWebsitePolicy.visibility_scope"}
+                            choices={resultsWebsiteVisibilityOptions()}
```
xalsina-sequent: Public access with area-based visibility is selectable here even though the server rejects it. We should add cross-field validation or disable the incompatible options.

packages/admin-portal/src/resources/Tally/TallyResults.tsx
```diff
@@ -27,229 +43,416 @@ interface TallyResultsProps {
...
         )
 
-        const resultsElection: Array<Sequent_Backend_Results_Election> | undefined = useMemo(
+        const elections = useMemo<Sequent_Backend_Election[]>(() => {
+            const electionById = new Map(
+                tallyData?.sequent_backend_election?.map((election) => [election.id, election]) ??
+                    []
+            )
+
+            return (tally?.election_ids ?? [])
```
xalsina-sequent, line 109: While tallyData is still loading—or if it fails—this produces an empty election list that is rendered as “No Result.” We should distinguish loading/error states from a genuinely empty result set.

packages/admin-portal/src/resources/ElectionEvent/EditElectionEventData.tsx
```diff
@@ -92,12 +115,38 @@ export const EditElectionEventData: React.FC = () => {
...
+                    variables: {
+                        election_event_id: electionEventId,
+                        status: policy.status,
+                        access: policy.access,
+                        visibility_scope: policy.visibility_scope,
+                    },
+                })
+            } catch (error) {
+                console.error(error)
+                notify("Failed to update the results website policy", {type: "error"})
```
xalsina-sequent: This policy mutation runs only after the event update commits. If it fails, the event changes remain saved while the policy does not. We should permission-gate these controls and avoid presenting both writes as one successful save.

packages/windmill/src/postgres/tally_results_publication.rs
```diff
@@ -0,0 +1,523 @@
...
+    tenant_id: &str,
+    election_event_id: &str,
+    route_scope: ResultsRouteScope,
+    route_election_id: Option<&str>,
+) -> Result<i32> {
+    let route_election_uuid = route_election_id.map(parse_uuid_v4).transpose()?;
+    let statement = tx
+        .prepare(
+            r#"
+                SELECT COALESCE(MAX(version), 0) + 1 AS version
```
xalsina-sequent, line 251: Possible race condition in version assignment. next_publication_version uses a plain SELECT COALESCE(MAX(version), 0) + 1 with no FOR UPDATE or advisory lock.

packages/ui-essentials/src/components/TallyResults/ResultsAndParticipation.tsx
```diff
@@ -0,0 +1,39 @@
...
+    processResults,
+}) => (
+    <Box className="seq-tally-results">
+        <ParticipationSummary
+            result={summary}
+            chartName={chartName}
+            labels={labels}
+            showWeight={showWeight}
+        />
+        {processResults ? (
```
xalsina-sequent: The old `counting_algorithm` guard deciding plurality-table vs IRV-table
rendering was removed; the new code falls back to the flat plurality-style
`CandidateResults` table for **any** algorithm whenever `processResults` is
null.
An INSTANT_RUNOFF contest is tallied but its
`process_results` annotation is still processing or fails to parse. The admin
Tally Results screen now renders a flat `cast_votes`/`winning_position` table
as if it were a plurality result, instead of showing nothing (the old,
guarded behavior) — misrepresenting how an IRV contest actually determined
its winner.

packages/voting-portal/src/routes/ElectionSelectionScreen.tsx
```diff
@@ -114,6 +115,18 @@ const isElectionEventEarlyVotingOpen = (electionEvent?: IElectionEvent): boolean
     )
 }
 
+const isResultsWebsiteEnabled = (electionEvent?: IElectionEvent): boolean => {
+    return electionEvent?.presentation?.results_website?.status === EResultsWebsiteStatus.ENABLED
```
xalsina-sequent: results_website is a JSON-encoded string, so reading .status directly always returns undefined and hides the results link. We should parse it with parseResultsWebsitePolicy before checking its status.

packages/admin-portal/src/resources/Tally/TallyResultsSectionGlobal.tsx
```diff
@@ -99,6 +122,74 @@ export const TallyResultsSectionGlobal: React.FC<TallyResultsGlobalCandidatesPro
...
+            totalVotesPercent: result.total_votes_percent,
+            totalValidVotes: result.total_valid_votes,
+            totalValidVotesPercent: result.total_valid_votes_percent,
+            totalInvalidVotes: result.total_invalid_votes,
+            totalInvalidVotesPercent: result.total_invalid_votes_percent,
+            explicitInvalidVotes: result.explicit_invalid_votes,
+            explicitInvalidVotesPercent: result.explicit_invalid_votes_percent,
+            implicitInvalidVotes: result.implicit_invalid_votes,
+            implicitInvalidVotesPercent: result.implicit_invalid_votes_percent,
+            blankVotes: result.total_blank_votes,
```
xalsina-sequent, line 144: Missing explicit/implicit blank-vote breakdown. Shouldn't we preserve explicit_blank_votes, implicit_blank_votes, and their percentages?

xalsina-sequent (changes requested):

PR #2871, opened 2026-07-19
🐞 Results Portal: Link from Voting Portal doesn't look good

Findeton:
Parent issue: https://github.com/sequentech/meta/issues/12576


Findeton:
recheck

PR #2877, opened 2026-07-21
✨ Smartlink for second gen (#2874)

Findeton:
Parent issue: https://github.com/sequentech/meta/issues/12275


xalsina-sequent (changes requested):
There are some issues that need to be addressed the others we can create a folloup.

Here is a review summary you can use as a prompt:

## Review summary — action-token-login-bridge + realm-attribute plumbing (step#2877)

### 🔴 Blocking

1. **Permission regression for existing tenants** — `update-realm-attributes` (harvest route + Hasura action) switched from `election-event-write` to the new `keycloak-realm-attributes-write` role. Only the **devcontainer** tenant import JSON was updated to grant it; I found no corresponding change to whatever provisions production/new tenant realms. Existing admins will silently lose the ability to change `voter-certificate-policy` until the new roles are backfilled. Needs a migration/provisioning story, or the route should accept either permission transitionally.
2. **Silent no-op editing in admin-portal** — in `EditElectionEventDataForm.tsx`, the write path bails out early when `!canEditRealmAttributes`, but the voter-certificate-policy `SelectInput` still renders as editable. A user with only `election-event-write` can change the dropdown, hit Save, and have the change silently dropped with no error shown. The control should be disabled instead.
3. **Update endpoint now accepts arbitrary realm attribute keys** — `apply_realm_attribute_updates` used to ignore unknown keys; now it writes **any** key (e.g. `acr.loa.map`, other extensions' config). That's a meaningful capability expansion of this endpoint, gated only by the new permission. If intentional, call it out explicitly in the docs/PR description; otherwise add an allow-list of known-prefix attributes.

### 🟡 Should address

- `x-hasura-election-event-id` claim was added to `JwtHasuraClaims` and is injected by the mapper, but nothing in either PR consumes it yet. If it's for a follow-up, say so in the PR description; otherwise it's scope creep in a security-sensitive claims mapper.
- `LoginBridge` logs the **serialized live action token** at debug (`log.debugf("baseUri: %s, tokenString: %s, ...")`). If debug logging is ever enabled in production, single-use login tokens land in log files. Since this PR is already curating this code (downgrading other lines from `infof`→`debugf`, redacting secrets elsewhere), the token string itself should be dropped or truncated here too.
- Build coupling: `Dockerfile.keycloak` now requires a BuildKit `additional_context beyond` to pull in `beyond`'s extension sources (see beyond#776's `pom.xml` `relativePath`) — a plain `docker build` without that named context fails. Worth documenting the required repo layout prominently, since it's currently only implicit.

### 🟢 Minor / nits

- `RealmNames.parseEventRealmName` splits on the *first* literal `-event-` segment — a tenant id that itself contains `-event-` would mis-parse. Fine for UUID tenants today; consider splitting on the *last* occurrence for robustness.
- `readPositiveLong`-style helpers use a boxed `Long` + null sentinel where `Optional<Long>` (or a small result type) would be more idiomatic.
- Inconsistent SPDX years (2025 vs 2026) across new files.

Findeton:
I will fix these issues in a follow up

PR #2928, opened 2026-08-03
🐞 Admin portal - Keycloak service account user counted as a elegible voter in the event dashboard

BelSequent:
Parent issue: https://github.com/sequentech/meta/issues/11175

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved voter and area filtering for requests with explicit or unspecified election and area criteria.
  * Areas without contests now remain visible when no election is selected.
  * Unassigned voters remain included in fully unfiltered results.
  * Service accounts are excluded from voter lists, lookups, and counts for more accurate results.
  * Filtering is now applied consistently across voter lists, totals, and paginated result counts.


packages/windmill/src/services/users.rs
```diff
@@ -1257,6 +1276,7 @@ pub async fn list_users_ids(
 
     debug!("parameters count: {}", next_param_number - 1);
     debug!("params {:?}", params);
+    let scope_clause = voter_scope_clause(&filters_clause);
```
eselimsen: I can't find any usages of this `list_users_ids` function, is it even used anywhere in the codebase? Even `grep`ing the whole project yields nothing (looks unused after 9.0 which have the only usage). Is there a machinery that still likely uses this, or shall we remove it?

packages/windmill/src/services/users.rs
```diff
@@ -802,6 +825,7 @@ pub async fn count_keycloak_users(
     };
 
     // Build the count query using only the necessary filtering clauses.
+    let scope_clause = voter_scope_clause(&filters_clause);
```
eselimsen: This `voter_scope_clause` helper looks extremely fragile to me. It hardcodes sql parameters within itself, if the call-site differs by parameter ordering in any way (including cherry-picking changes to other branches) it will cause subtle bugs. I think, it should take those as function parameters, at least.

eselimsen (changes requested):
It looks good to me, just a couple of quick pointers.

eselimsen (approved):
💯

PR #2931, opened 2026-08-04
🐞 IVR: Error on round trip

BelSequent:
Parent issue: https://github.com/sequentech/meta/issues/12712

### DIFF
--- a/packages/admin-portal/src/services/generated/ivr_emulator_wasm.d.ts
+++ b/packages/admin-portal/src/services/generated/ivr_emulator_wasm.d.ts
@@ -62,10 +62,9 @@ export interface InitOutput {
     readonly ivremulatordriver_new: (a: any) => [number, number, number];
     readonly ivremulatordriver_send_input: (a: number, b: number, c: number) => void;
     readonly ivremulatordriver_send_timeout: (a: number) => void;
-    readonly ring_core_0_17_14__bn_mul_mont: (a: number, b: number, c: number, d: number, e: number, f: number) => void;
-    readonly wasm_bindgen__closure__destroy__h7baa5a84311a3b04: (a: number, b: number) => void;
-    readonly wasm_bindgen__convert__closures_____invoke__h5fed3dcc0d3dcd7b: (a: number, b: number, c: any) => [number, number];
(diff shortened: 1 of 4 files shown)


packages/admin-portal/src/services/generated/ivr_emulator_wasm.d.ts
```diff
@@ -62,10 +62,9 @@ export interface InitOutput {
     readonly ivremulatordriver_new: (a: any) => [number, number, number];
     readonly ivremulatordriver_send_input: (a: number, b: number, c: number) => void;
     readonly ivremulatordriver_send_timeout: (a: number) => void;
-    readonly ring_core_0_17_14__bn_mul_mont: (a: number, b: number, c: number, d: number, e: number, f: number) => void;
-    readonly wasm_bindgen__closure__destroy__h7baa5a84311a3b04: (a: number, b: number) => void;
-    readonly wasm_bindgen__convert__closures_____invoke__h5fed3dcc0d3dcd7b: (a: number, b: number, c: any) => [number, number];
-    readonly wasm_bindgen__convert__closures_____invoke__h331bfc0ec3639989: (a: number, b: number, c: any, d: any) => void;
+    readonly wasm_bindgen__closure__destroy__hd27b649bb08bb176: (a: number, b: number) => void;
```
Findeton, line 65: how are you guys generating this file? shouldn't this be autogenerated?

BelSequent: Yes its autogenerated

eselimsen: This file is auto-generated, but this section is not used at all. Notice that this is a typescript definition file, so it is not interpreted for the execution, but only provides typing support for our emulator wasm interface. This section is wasm-bindgen internals that change quite often, but we're not affected by that.

The parts we use are between:
https://github.com/sequentech/step/blob/41c6f3881861f9111082b796c65c9a3e95e31277/packages/admin-portal/src/services/generated/ivr_emulator_wasm.d.ts#L3
https://github.com/sequentech/step/blob/41c6f3881861f9111082b796c65c9a3e95e31277/packages/admin-portal/src/services/generated/ivr_emulator_wasm.d.ts#L52

Perhaps we can create an automation in the future to strip off these parts.

PR #2958, opened 2026-08-06
✨ Unify election event generation, validation and import on one shared Rust core

edulix:

Companion to https://github.com/sequentech/beyond/pull/802. Beyond's Rust CI looks
for a step branch of the same name, and this is it.

Unifies the three tools that build or check an election event import onto one
shared Rust definition. This PR is the step half; it lands in small increments.

## Why here and not in beyond

The dependency between the repositories runs one way. `beyond` is a **git
submodule of step** and its Rust workspace path-depends on step's crates
(`sequent-core = { path = "../../packages/sequent-core" }`). **step cannot depend
on beyond**, so anything windmill must use has to live here. A shared crate in
`beyond/packages` would serve the two configuration SPAs and leave the server on
its own implementation — which is the problem, not the fix.

`sequent-core` needs no new mechanism for the browser half either: it already
compiles to WASM (`build_wasm.yml`, `wasm-pack build --target web`) and that build
is already vendored into admin-portal, voting-portal, ui-core and ballot-verifier
as `file:./rust/sequent-core-0.1.0.tgz`.

## What is here so far

**`sequent-core::election_config`** — a new module for the import bundle format.
Gated on `default_features`, matching `types::hasura` whose entities the schema
will be built from; the WASM build enables that feature, so the module is present
in the browser.

**The report types move in.** `Report`, `ReportCronConfig`, `ReportType` and
`EReportEncryption` come from `windmill::postgres::reports` and
`windmill::services::reports::template_renderer`, so that the tools which *write*
an import describe reports the same way the importer reads them. windmill
re-exports all four, so its ~10 call sites are untouched — net −44 lines.

`ReportWrapper` and its `TryFrom<Row>` deliberately stay in windmill: they need
`tokio_postgres`, which has no place in a module that has to compile to WASM.

Six unit tests cover the wire forms that are part of the file format rather than
implementation details — the `snake_case` encryption policy the reports CSV
carries, a cron config surviving `{}`, and `permission_label` being a list here
where `Election`'s is a string.

## Verified

- `cargo check -p sequent-core` and `cargo check -p windmill` both clean.
- `cargo test -p sequent-core --lib --features default_features election_config` — 6 passed.

## The stack above this PR

Each is its own draft PR, based on the one before. Review bottom-up.

1. **this PR** — the module, and the report types moved in.
2. https://github.com/sequentech/step/pull/2961 — `election_config::schema`, the bundle schema moved out of windmill.
3. https://github.com/sequentech/step/pull/2962 — `election_config::validate`, the pure checks returning structured problems.
4. https://github.com/sequentech/step/pull/2968 — windmill validating before importing.
5. https://github.com/sequentech/step/pull/2970 — `election_config::emit`, the bundle's file formats written once.
6. https://github.com/sequentech/step/pull/2971 — `election_config::paths`, how a spreadsheet cell becomes JSON.
7. https://github.com/sequentech/step/pull/2972 — `election_config::sheet` and `::xlsx`, reading an authoring workbook.
8. https://github.com/sequentech/step/pull/2973 — `election_config::ids`, the deterministic uuid5 factory.
9. https://github.com/sequentech/step/pull/2974 — `election_config::render`, the base entity templates.
10. https://github.com/sequentech/step/pull/2975 — `election_config::build`, the bundle's JSON document.
11. https://github.com/sequentech/step/pull/2976 — the bundle's CSV members.
12. https://github.com/sequentech/step/pull/2977 — the auth presets and the Keycloak realm.
13. https://github.com/sequentech/step/pull/2978 — the files a bundle ships as, and the zip.
14. https://github.com/sequentech/step/pull/2979 — janitor folded into `step-cli`, and proven byte-identical to it.
15. https://github.com/sequentech/beyond/pull/803 — retires `beyond/packages/janitor` and points its docs at `step-cli` (in beyond, stacked on #802).
16. https://github.com/sequentech/step/pull/2980 — `election_config::wasm`, the browser's view of the same checks.
17. _next_ — the two SPAs in `beyond/packages`, and a WASM package that enables the config features for them.

## Screenshots/videos Before

`<empty>`

`<empty>` — no user-visible change in this PR; it is a refactor.

### DIFF
--- a/packages/sequent-core/src/election_config/mod.rs
+++ b/packages/sequent-core/src/election_config/mod.rs
@@ -1,0 +1,28 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! The election event import bundle: what it contains, and whether it is valid.
+//!
+//! This is the single definition of the format shared by everything that
+//! produces or consumes an import — windmill's importer on the server, and the
+//! configuration tools in `beyond/packages` in the browser.
+//!
+//! It lives here rather than in `beyond` because the dependency between the
+//! repositories runs one way: `beyond` is a git submodule of step and
+//! path-depends on this crate, so step cannot depend on `beyond`. Anything
+//! windmill must use has to be here. `sequent-core` also already compiles to
+//! WASM and is already vendored into the front ends as a package, so both
+//! consumers reach this module through paths that already carry production
+//! code.
+//!
+//! Everything in this module must stay **pure** — no database, no IO, no
+//! clock — or it cannot run in a browser, and the browser is where a delivery
+//! engineer wants the answer.
+//!
+//! See `beyond/docs/docusaurus/docs/engineering/election-config-architecture.md`
+//! and <https://github.com/sequentech/meta/issues/12769>.
+
+pub mod report;
+
+pub use report::{EReportEncryption, Report, ReportCronConfig, ReportType};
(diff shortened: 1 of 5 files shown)


edulix:

PR #2961, opened 2026-08-06
♻️ Move the import bundle schema into sequent-core::election_config

edulix:

**Stacked on https://github.com/sequentech/step/pull/2958.** Base branch is
`feat/meta-12769-unified-election-config/main`, so this diff shows only what is
added on top.

Moves `ImportElectionEventSchema` out of windmill into
`sequent-core::election_config::schema`, so the tools that *write* an import
bundle describe it the same way the importer reads it. Until now each reproduced
it: janitor in Handlebars templates, the Election Architect in a hand-built
TypeScript object that was not even the importable shape.

windmill re-exports it, so its own call sites are unchanged.

## Two field types differ, both to keep the module WASM-compatible

| Field | Was | Now | Why |
|---|---|---|---|
| `tenant_id` | `Uuid` | `String` | Making it a `Uuid` pulls that crate into `default_features` and `getrandom` into the WASM build. Import replaces this value with the importing request's tenant regardless of what it says, and every use in windmill stringified it already. |
| `keycloak_event_realm` | `RealmRepresentation` | `serde_json::Value` | That type comes from the `keycloak` crate, which pulls `reqwest`. |

Neither loses anything real:

- The export path still runs `parse_uuid_v4`, so an export cannot emit a malformed
  tenant id. Validation will report one as a readable problem where a `Uuid` field
  produced an opaque serde error.
- serde round-trips the realm exactly — a test asserts that, including keys this
  crate has never heard of. windmill deserializes it into the typed form at the one
  place it talks to Keycloak. Validating a realm needs a live Keycloak anyway, so
  it was never something the shared validation could check.

## Tests

Five new, covering what the format guarantees rather than the struct's shape:

- a minimal bundle deserializes;
- a missing `version` falls back to the historical default, so bundles predating
  the field still import;
- the realm round-trips untouched, unknown keys included;
- a missing non-`Option` field is rejected — the failure the shared validation
  exists to pre-empt;
- `tenant_id` survives unaltered so `replace_ids` can map it.

The missing-field test builds its input by removing the key from the parsed tree
rather than by editing the JSON text: a string replacement that stopped matching
would silently pass instead of failing, which is exactly what happened on the
first attempt.

## Verified

- `cargo check -p sequent-core --features default_features` and `cargo check -p windmill` clean.
- `cargo test -p sequent-core --lib --features default_features election_config` — 11 passed.
- `cargo fmt --check` clean on both crates.

## Screenshots/videos Before

`<empty>`

`<empty>` — refactor, no user-visible change.

### DIFF
--- a/packages/sequent-core/src/election_config/mod.rs
+++ b/packages/sequent-core/src/election_config/mod.rs
@@ -24,5 +24,7 @@
 //! and <https://github.com/sequentech/meta/issues/12769>.
 
 pub mod report;
+pub mod schema;
 
 pub use report::{EReportEncryption, Report, ReportCronConfig, ReportType};
+pub use schema::ImportElectionEventSchema;
--- a/packages/sequent-core/src/election_config/schema.rs
+++ b/packages/sequent-core/src/election_config/schema.rs
@@ -1,0 +1,184 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! The election event import bundle.
+//!
+//! This is the shape of `export_election_event-<uuid>.json`, and the single
+//! definition of it. It was previously declared in
+//! `windmill::services::import::import_election_event`, which meant every tool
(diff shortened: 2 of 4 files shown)


edulix:

PR #2962, opened 2026-08-06
✨ Add shared election config validation to sequent-core

edulix:

**Stacked on https://github.com/sequentech/step/pull/2961.** Base branch is
`feat/meta-12769-schema/main`, so this diff shows only what is added on top.

The pure half of checking an import bundle — no database, no IO, no clock. That is
what lets the same code answer in a browser before an upload and on the server
before a transaction. Anything needing the database (does this tenant exist, is
this area name taken) stays in windmill on top of this pass.

```rust
let report = sequent_core::election_config::validate(&bundle);
if report.has_errors() { … }
```

## What a Problem carries

| Field | Why |
|---|---|
| `code` | Machine-readable, so a front end can group, translate or link without matching English |
| `severity` | Error or Warning — see below |
| `path` | `contests[2].max_votes`, dotted into the bundle |
| `external_id` | The UUIDs are regenerated on import and mean nothing to whoever fixes the source; the `external_id` is what they typed |

Never stops at the first problem: fixing a configuration one error per run is
miserable.

## Warnings are not lesser errors

A warning says the bundle is self-consistent but probably not what its author
meant. The one that exists today is the expensive one: **a permission label hides
an entity from every administrator who does not hold it**, so an event imports
cleanly and then lists nothing. That is exactly what happened on the first real
SEIU import. The bundle cannot know who holds which label, so it cannot be an
error — but it must be said.

## The rules

From what the importer rejects, and from what janitor learned the hard way. The
second kind matters more: those bundles satisfy every type in the schema and still
fail on election day.

- **Identity** — tenant id present and UUID-shaped, event id and encryption
  protocol present, at least one election and one area.
- **References** — contests to elections, candidates to contests, area/contest
  links both ways.
- **Area tree** — parents exist, no cycles. An infinite tree hangs the Admin
  Portal rather than failing the import.
- **Contest arithmetic** — `min_votes` ≤ `max_votes`, winners and selections
  within the candidate count, vote counts present.
- **Tally system** — `voting_type` and `counting_algorithm` must agree. Ballot
  encoding follows the *algorithm*, so a preferential contest counted by
  `plurality-at-large` imports cleanly and then reads a voter's rankings as
  unordered selections.
- **Ballot coverage** — no contest on nobody's ballot, no leaf area with an empty
  one. Neither breaks the import; both mean somebody's vote is missing.
- **Permission labels** — the warning above.
- **Identifiers** — unique across every collection, not just within one. A contest
  and an area colliding is just as bad as two contests colliding.

`tenant_id`'s format is checked here rather than by its type — what the schema
traded away to stay WASM-safe. `looks_like_uuid` checks the shape rather than
pulling the `uuid` crate, and therefore `getrandom`, into the WASM build to inspect
a string.

## Tests — 50

Each starts from one bundle that validates cleanly and breaks exactly one thing, so
a failure names the rule rather than producing a pile of unrelated problems. If the
fixture ever stops validating, `a_sound_bundle_has_no_errors` fails first and says
why.

Two are about the failure modes of the checks themselves rather than of a bundle:

- a negative `winning_candidates_num` must not wrap into a huge `usize` and
  silently pass — the first draft had `winners as usize > available`;
- every algorithm in each list is exercised against the voting type it belongs
  with, rather than a sampled few.

## Verified

- `cargo check -p sequent-core --features default_features` and `-p windmill` clean.
- `cargo test -p sequent-core --lib --features default_features election_config` — 50 passed.
- `cargo fmt --check` clean.

## Next on the stack

windmill calling this before its transaction, so the server and the browser reach
the same verdict from the same code.

## Screenshots/videos Before

`<empty>`

`<empty>` — no user-visible change yet; nothing calls this until the next PR.

### DIFF
--- a/packages/sequent-core/src/election_config/mod.rs
+++ b/packages/sequent-core/src/election_config/mod.rs
@@ -23,8 +23,15 @@
 //! See `beyond/docs/docusaurus/docs/engineering/election-config-architecture.md`
 //! and <https://github.com/sequentech/meta/issues/12769>.
 
+pub mod problem;
 pub mod report;
 pub mod schema;
+pub mod validate;
 
+#[cfg(test)]
+mod validate_tests;
+
+pub use problem::{Code, Problem, Report as ValidationReport, Severity};
 pub use report::{EReportEncryption, Report, ReportCronConfig, ReportType};
 pub use schema::ImportElectionEventSchema;
+pub use validate::validate;
--- a/packages/sequent-core/src/election_config/problem.rs
+++ b/packages/sequent-core/src/election_config/problem.rs
@@ -1,0 +1,237 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! What validation reports.
+//!
+//! Structured rather than a formatted string, because the same problem has to be
+//! rendered three ways: a line in `step-cli`'s output, a row in a browser's
+//! problem list, and an error from a server-side import. Each wants the pieces
+//! arranged differently, and only the caller knows which.
+//!
+//! Every problem carries a machine-readable [`Code`] so a front end can group,
+//! translate or link them without matching on English text.
+
+use serde::{Deserialize, Serialize};
+use std::fmt;
+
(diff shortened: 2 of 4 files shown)


edulix:

PR #2960, opened 2026-08-06
🌐 i18n: plurals, dead translations, language defects, NIST terminology and admin translations (stack)

edulix:
**Integration PR for the i18n stack — squash-merge target.**

The five PRs below exist so the work can be *reviewed* in coherent pieces. This one exists so it can be
*merged* in one go: it carries the whole stack against `main`, so squashing it lands a single commit instead of
five separate merges and four base-branch retargets.

Review in the child PRs. Merge here.

| # | change | commits |
|---|---|---|
| [#2952](https://github.com/sequentech/step/pull/2952) | pluralise voter-facing counts with i18next, shorten ballot warnings, document the convention | 4 |
| [#2953](https://github.com/sequentech/step/pull/2953) | delete the unreferenced `ui-essentials` translations and the tutorial section maintaining them | 1 |
| [#2954](https://github.com/sequentech/step/pull/2954) | fix broken placeholders, inconsistent terminology, and the untranslated Dutch bundle | 2 |
| [#2956](https://github.com/sequentech/step/pull/2956) | unify election terminology against the NIST glossary | 3 |
| [#2957](https://github.com/sequentech/step/pull/2957) | translate the admin-portal strings still left in English | 3 |

Tracked by [meta#12764](https://github.com/sequentech/meta/issues/12764),
[#12765](https://github.com/sequentech/meta/issues/12765),
[#12766](https://github.com/sequentech/meta/issues/12766),
[#12768](https://github.com/sequentech/meta/issues/12768),
[#12770](https://github.com/sequentech/meta/issues/12770).

## What lands, in one paragraph

The frontends had no agreed vocabulary and no working plural support. Counts were pluralised by hand with
`count === 1` ternaries and a dead i18next-v3 `_plural` key; the same concept was named three or four different
ways per language, and several of those names were the wrong word — `concurso`/`concours` (a prize
competition), Dutch `verkiezing` (an *election*, a level above a contest), `wedstrijd` (a sports match). Eight
translation files nothing loaded were still being maintained by a tutorial that pointed at a file which does not
exist. This stack fixes the plural mechanism, deletes the dead weight, repairs the language defects, settles one
term per language against NIST's glossary, and translates the 516 admin-portal strings still sitting in English.

## Defects a voter or operator would have hit

- **Galician audit screen had a dead link** — the one screen where a voter confirms their ballot was encrypted
  correctly rendered `href="{{linkToBallotVerifier}}"` literally.
- **Basque ballot locator printed a raw `{{ballotId}}`.**
[...]

PR #2963, opened 2026-08-06
✨ i18n: add Spanish and Catalan informal (tú/tu) register variants

edulix:
Parent issue: https://github.com/sequentech/meta/issues/12764 (Part 6)

> **Stacked on [#2957](https://github.com/sequentech/step/pull/2957)** → [#2956](https://github.com/sequentech/step/pull/2956) → [#2954](https://github.com/sequentech/step/pull/2954) → [#2953](https://github.com/sequentech/step/pull/2953) → [#2952](https://github.com/sequentech/step/pull/2952).
> Base branch is `fix/meta-12769-admin-untranslated/main`. Squash-merge through the integration PR, [#2960](https://github.com/sequentech/step/pull/2960).

## What this changes

Four selectable locales instead of two:

| code | register | `<html lang>` |
|---|---|---|
| `es` | formal — *usted* | `es` |
| `es-tu` | informal — *tú* | `es-x-tu` |
| `cat` | formal — *vós* | `ca` |
| `cat-tu` | informal — *tu* | `ca-x-tu` |

**Plain `es`/`cat` are unchanged and stay formal**, so no existing deployment shifts register on upgrade. A
deployment opts in by adding `es-tu`/`cat-tu` to `enabled_language_codes`.

## The problem was not that the register was wrong — it was that there were two

Both languages addressed the voter two ways at once, inside single sentences. Catalan
`reviewScreen.description`:

```
Per fer canvis a les seves seleccions, faci clic al botó "<b>Edita el teu vot</b>", …
```

`les seves` and `faci` are formal; `el teu` in the button label is not. Spanish `reviewScreen.error.*`
alternated key by key within one error family — `inténtelo`/`contacte` for sixteen keys, `inténtalo`/`contacta`
for the next fifteen, then back.

Over the four voter-facing packages, 452 strings per language:

| | formal only | informal only | both in one string |
|---|---|---|---|
| es | 74 | 49 | 5 |
| cat | 56 | 65 | 7 |

Neither register won, which is why "pick one" was never the fix. Different customers want different ones: an
institutional election wants *usted*, a membership or student vote often wants *tú*. So the register became a
property of the locale.

## Hand-written, and here is why

**309 string pairs written by hand.** A rule-based `usted`↔`tú` converter was built first and thrown away. It
looked fine — 128 of 131 Spanish strings passed a marker check — until the output was read:

| input | what the rules produced | why it is wrong |
|---|---|---|
| `cuando el votante elige auditar` | `votante elija` | third person, not an imperative |
| `Error de tiempo de espera` | `tiempo de espere` | `espera` is a noun here |
| `Esta pantalla confirma que…` | `pantalla confirme` | the screen confirms, not you |
| `usted es elegible` | `tú es elegible` | needs *eres* |
| `Tu voto es inválido` | `Tu voto eres inválido` | the *vote* is invalid |
| `Esta acción no se puede deshacer` | `no se puedes deshacer` | the *action* cannot |
| `El orden de preferencia tiene huecos` | `preferencia tienes huecos` | the *order* has |

Every rule added to fix one case exposed another — the distinction is syntactic, not lexical. For a voting
product a mangled ballot instruction reaching a voter is worse than shipping one register, so the whole set was
written out by hand instead.

The 9 strings that remain byte-identical between the registers are genuinely register-free — `se usa`,
`le permite al votante`, `La impresión está desactivada` — and a test asserts the rest actually differ.

## Design

**Complete bundles, not overlays.** `es-tu.ts` carries all 452 strings, of which 149 differ from `es.ts`. An
overlay would be smaller, but a key present in the base and missing from the variant would drop the voter back
into the other register mid-screen. A test asserts the key sets are identical. `fallbackLng` still points each
variant at its base rather than English, so a key added to a base and not yet mirrored degrades to the right
language instead of to English.

**`language` is overridden explicitly.** `SettingsLanguages` renders `t("language", {lng})` for every bundle in
the resource store — without this an operator would see two entries both reading "Español". They read
"Español (tú)" and "Català (tu)".

**Register rides in a private-use subtag.** BCP 47 has no registered subtag for register, so `<html lang>` gets
`es-x-tu`: well-formed, and assistive technology falls back to Spanish rather than rejecting an unregistered
variant.

## One bug this exposed

[`locale.rs`](https://github.com/sequentech/step/blob/main/packages/sequent-core/src/util/locale.rs) did
`lang.split('-').next()` — discarding every subtag after the first. Harmless while every internal code is a bare
primary tag, but it means **no** locale variant can ever round-trip: `es-x-tu` silently normalises to `es`. Fixed,
with tests covering the round trip for all ten codes and asserting a region is never mistaken for a register
(`es-MX`, `ca-ES-valencia`).

The marker is split off in TypeScript rather than inside sequent-core, because the WASM bindings ship as a
**committed tarball** whose sha1 is pinned in four `yarn.lock` entries and is regenerated only through Nix +
`wasm-pack`. Rebuilding it is out of scope here; `locale.rs` implements the same rule, so the two agree once
that artifact is next rebuilt. `registerLocale.ts` holds the pure string handling so it is testable without the
WASM import.

## Finding the leaks

A register variant is only worth having if each bundle is consistent, and the strings that break that rule are
the hard ones to see: a single imperative in the wrong form, surrounded by correct ones.

A list of verb pairs does not find them — it only ever finds what it already knows. What works is deriving the
lexicon from the bundles. Each formal string is aligned with its `-tu` twin word by word, and the substitutions
that look like the imperative inversion — same stem, ending swapped — are kept. The corpus yields its own
lexicon that way (`guarde→guarda`, `ingrese→ingresa`, `elija→elige`, `arrastre→arrastra`, …), which is then
used to flag any formal form sitting in a `-tu` bundle or vice versa.

Two refinements matter. Forms ending in `-s` are dropped, because the 2nd-person indicative is lexically
identical to the ordinary 3rd person — "la impresión **está** desactivada" is correct in both registers — which
is the same ambiguity that makes rule-based conversion unsafe. And a separate pass flags pairs that are
byte-identical where the *other* language's pair differs, which catches conversions skipped altogether; a
lexicon cannot see those, and two of the leaks below were found only this way.

Six strings were addressing the voter in the wrong register:

| file | was | now |
|---|---|---|
| `voting-portal/es.ts` `INCONSISTENT_HASH` | `informa` (tú) | `informe` |
| `voting-portal/cat.ts` `INCONSISTENT_HASH` | `informa` (tu) | `informeu` |
| `voting-portal/cat.ts` `errorDialogPrintBallotReceipt` | `intenti` (vostè) | `intenteu-ho` |
| `ui-core/cat.ts` `importErrorDescription` | `Vas triar` (tu) | `Vau triar` |
| `ballot-verifier/cat.ts` `404.subtitle` | `busques` (tu) | `busqueu` |
| `voting-portal/cat.ts` `warningDialog.continue` | `Continua` (tu) | `Continueu` |

The Catalan ones are all *vós* bundles: `cat.ts` uses vós over vostè by 123 forms to 5, and `vostre` over
`el seu` by 63 to 1.

`errorDialogPrintBallotReceipt` also turned out to be the single string in each of the four bundles that
skipped the `"<statement>. <please>, <imperative>."` punctuation the other ~50 follow, and the Catalan informal
copy had drifted lexically from its twin ("Hi ha hagut" for "Ha ocorregut") for no register reason. All four
now differ only in the address form.

## Deliberately not done

- **`admin-portal` has no variant.** Admin copy addresses operators, not voters. It inherits the ui-core variant
  and falls back to `es`/`cat` for its own strings.
- **`getAllLangs()` is untouched** — it seeds per-election *content* translations (candidate names), which have
  no register.
- **No variants for gl, eu, nl, tl, fr.** French *vous*/*tu* is a real distinction, but nobody has asked for it.

## Verification

```
cargo test (sequent-core)     132 passed — incl. 10 locale tests, round trip + region-is-not-a-register
placeholder parity vs en      0 mismatches across 16,424 strings
variant vs base               922 keys compared, 0 missing, 0 extra, 0 placeholder mismatches
register consistency          6 leaks found and fixed; 0 remaining, 0 unconverted pairs
plural parity                 104 `_many`/`_other` pairs identical, 0 missing
yarn prettify                 All matched files use Prettier code style! (6/6)
yarn lint                     clean
ui-core                       66 tests passed (12 new)
ui-essentials                 14 · voting-portal 26 · admin-portal 50 · results-portal 7
```

`Valencià` vs `Català`: `ui-core/cat.ts` labels the language "Valencià" while `results-portal/cat.ts` says
"Català". That predates this PR and is left as-is in case it is deliberate for a Valencian deployment — the
variants inherit whichever label their base uses.


edulix:
`Run Rust tests (velvet)` is red here, and it is not this change.

The failure is `Error obtaining the browser / Unable to make method calls because underlying connection is closed` inside `print_to_pdf` — the job's own retry loop (`attempt 2/5`) also fails. The velvet CLI suite renders PDFs through headless Chrome.

Checked rather than assumed:

- The commit that triggered it changes **30 TypeScript files and 0 Rust files**.
- velvet depends on `sequent-core`, but my only Rust change in this stack is `sequent-core/src/util/locale.rs`, a pure string function; velvet never references `locale`.
- Run locally on this branch: **9 failed**, all with the same headless-Chrome error.
- Run locally on the parent branch `fix/meta-12769-admin-untranslated/main`, which CI reports **green** for this same job: **the identical 9 tests fail**, same error.

So the suite is environment-dependent and its CI result reflects browser availability rather than the diff. #2963 and #2964 share a head commit and run this Chrome-heavy job concurrently, which is likely why the tip trips it while the parents do not.

Nothing to fix in this PR. Flagging it because a suite whose red/green is luck is worth knowing about — happy to open a separate issue for making the PDF-rendering tests resilient or gating them behind a browser-available check.

PR #2968, opened 2026-08-06
✨ Validate an import bundle with the shared rules before importing it

edulix:

**Stacked on https://github.com/sequentech/step/pull/2962.** Base branch is
`feat/meta-12769-validate/main`, so this diff shows only what is added on top.

Makes the shared validation load-bearing: windmill runs it on the bundle before
importing, and refuses if anything fatal turns up. The operator gets every problem
at once, in the same wording the browser-side tools will show before an upload.

Validated as written, before `replace_ids` rewrites the identifiers — a problem
naming an id the author never chose is not much use to them.

Additive: the checks that follow need the database, and this pass by design does
not touch it. `check_only` is unaffected.

## The part that deserves scrutiny: ballot coverage is now a warning

Wiring validation into the import path means **a bundle that used to import can
now be refused** — including the platform's own export, re-imported for disaster
recovery.

Three rules were fatal and should not have been:

- a contest with no candidates yet,
- a contest not yet on any area's ballot,
- an area nobody has assigned contests to.

An event still being configured legitimately looks like that. Those bundles are
entirely self-consistent and import into a working event; they just mean nobody
votes there yet. Refusing them would break recovery in order to enforce a rule
about authoring, so they are warnings.

The line is now:

| | |
|---|---|
| **Error** | The bundle is internally inconsistent, or would be rejected or corrupted by the platform |
| **Warning** | The bundle is consistent, but the configuration looks unintended |

Authoring tools should treat warnings as blocking — `step-cli` and the SPAs will do
that with a strict mode, rather than by lying about the severity here. Two tests
pin the line from both sides: `an_event_still_being_configured_can_be_re_imported`
and `an_inconsistent_bundle_is_still_refused`.

## Checked against real data

Adds a `validate_bundle` example — the same call `step-cli` and the browser will
make — for checking an export from a shell:

```bash
cargo run -p sequent-core --features default_features --example validate_bundle -- export_election_event-….json
```

Run against the generated SEIU1000 bundle:

```
warning: elections[].permission_label: permission labels in use: statewide-officers,
dlc-officers-dburs, cburs. Anything carrying a label is hidden from every
administrator without it…
0 error(s), 1 warning(s)
```

That warning is the thing that made every election invisible on the first real
import of that event. The Rust validator and janitor's independently written
Python rules reach the same verdict on the same real bundle — which is the point of
sharing them.

## Verified

- `cargo test -p sequent-core --lib --features default_features election_config` — 52 passed.
- `cargo fmt --check` clean on both crates.

## Screenshots/videos Before

`<empty>`

_To be added — the failure message an operator sees on a rejected import._

### DIFF
--- a/packages/sequent-core/examples/validate_bundle.rs
+++ b/packages/sequent-core/examples/validate_bundle.rs
@@ -1,0 +1,32 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Validate an election event bundle from a file.
+//!
+//! `cargo run -p sequent-core --features default_features --example validate_bundle -- <path>`
+//!
+//! A thin harness for checking a real export against the shared rules; the same
+//! call step-cli and the browser make.
+
+use sequent_core::election_config::{validate, ImportElectionEventSchema};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let path = std::env::args()
+        .nth(1)
+        .ok_or("usage: validate_bundle <path>")?;
+    let text = std::fs::read_to_string(&path)?;
+    let bundle: ImportElectionEventSchema = serde_json::from_str(&text)?;
+
+    let report = validate(&bundle);
+    print!("{report}");
+    println!(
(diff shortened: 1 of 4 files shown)


edulix:

PR #2970, opened 2026-08-06
✨ Share the writers for the bundle's own file formats

edulix:

**Stacked on https://github.com/sequentech/step/pull/2968.** Base branch is
`feat/meta-12769-windmill-validate/main`, so this diff shows only what is added on
top.

The zip the Admin Portal accepts holds one JSON document and up to three CSVs, and
two of those CSVs are read **positionally** by the importer. That byte shape was
implemented three times — windmill's exporter, janitor's Python, the Election
Architect's TypeScript — and the copies had drifted.

`election_config::emit` is the single implementation. Pure, like the rest of the
module: strings and byte vectors, no filesystem, so the same code writes a file
from `step-cli` and offers a download from a browser.

## The parts worth having in one place

**JSON-in-CSV.** Every field holds a JSON literal which is then CSV-quoted on top,
so a string arrives wrapped in three double quotes:

```
id,labels,cron_config
"""2c978b94-…""",null,"{""cron"":null,""scheduled_date"":""2026-10-24T16:15:00.000Z""}"
```

It looks like an escaping bug and is not — the importer parses each field with
`deserialize_str` after CSV decoding, so the JSON layer is load-bearing.

**A SQL NULL is written bare and unquoted.** That is why `JsonField` distinguishes
`Null` from `Value(Value::Null)`. Fold them together and an empty column becomes
indistinguishable from one holding the JSON value `null`.

**Only a *lone* empty field needs quoting.** In a single-column row an unquoted
empty field is a blank line, which a reader cannot tell from no row at all. An
empty field beside others is unambiguous and stays bare — which is also what
Python's `csv` writer does. So the special case lives in `join_csv`, which knows
the row length, and not in `escape_csv`, which does not.

**The column orders are the format.** `SCHEDULED_EVENT_COLUMNS` and
`REPORT_COLUMNS` are read by index: `import_scheduled_events.rs` takes the payload
at 10, `process_reports_file` takes `election_id` at 1 and `permission_label` at 7.
Naming them puts the ordering somewhere a reader can find it.

**`MULTI_VALUE_SEPARATOR` is a single pipe**, matching the importer. Workbooks use
`||`, so whichever tool reads one has to convert; a `||` that reaches a CSV is read
as one value containing an empty one.

Line endings are `\n`. `\r\n` is the `csv` crate's default and the reader accepts
either, but Windows endings diff badly in review for no benefit.

## Checked against real output

`it_matches_what_janitor_already_writes` pins the property that justifies replacing
three implementations with one: the expected row is copied verbatim from the file
janitor's Python emitted for the SEIU1000 event, and Rust reproduces it byte for
byte.

## Verified

- `cargo test -p sequent-core --lib --features default_features election_config` — **72 passed**, 0 failed (52 before, 20 new).

## Still to come on this stack

The duplicate writers are not deleted yet — windmill's exporter and janitor's
Python both stay until `step-cli` reads workbooks through this core, so each
replacement lands with something exercising it.

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/sequent-core/src/election_config/emit.rs
+++ b/packages/sequent-core/src/election_config/emit.rs
@@ -1,0 +1,487 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Writing the files an election event import is made of.
+//!
+//! The zip the Admin Portal accepts holds one JSON document and up to three
+//! CSVs, and two of those CSVs are read **positionally** by the importer, with a
+//! byte shape that is easy to get subtly wrong. Before this module that shape was
+//! implemented three times — in windmill's exporter, in janitor's Python, and in
+//! the Election Architect's TypeScript — and the copies had drifted.
+//!
+//! Pure, like the rest of this module: these functions return strings and byte
+//! vectors. Nothing here touches the filesystem, so the same code writes a file
+//! from `step-cli` and offers a download from a browser.
+
+use serde::Serialize;
+use serde_json::Value;
+
+/// Separator for multi-valued Keycloak user attributes.
+///
+/// A single pipe, matching
+/// `crate::services::keycloak::user::MULTIVALUE_USER_ATTRIBUTE_SEPARATOR`. The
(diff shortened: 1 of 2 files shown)


edulix:

PR #2971, opened 2026-08-06
✨ Share how a spreadsheet cell becomes JSON

edulix:

**Stacked on https://github.com/sequentech/step/pull/2970.** Base branch is
`feat/meta-12769-emit/main`, so this diff shows only what is added on top.

An authoring spreadsheet's column headers are paths into the target JSON:
`presentation.i18n.en.name` means nested objects four deep. That mapping is what
makes a workbook client-agnostic — a new column lands in the output with no code
change — and until now it existed only in janitor's Python.

`election_config::paths` is the Rust that `step-cli` and a browser need in order to
read the same workbook the same way. Ported rule for rule, then tested harder than
the original: **27 tests**, several for cases the Python has never been asked
about.

## The distinctions worth naming

**A blank cell and a cell holding `null` are different things**, and stay
different all the way through. Blank means the author said nothing, so the
template default holds; `null` means they cleared it. `coerce_cell` returns
`Option<Value>` for exactly that reason — `Some(Value::Null)` is not `None`.
Collapse them and there is no way to remove a default.

**Multi-value is a property of the column, not the content.** A cell with no
separator in it still becomes a one-element list, or the JSON type emitted would
depend on the data. Note `paths::MULTI_VALUE_SEPARATOR` is `||`, the workbook's,
while `emit::MULTI_VALUE_SEPARATOR` is `|`, the importer's. Two constants because
they are two formats; something has to convert, and a `||` that reaches a CSV is
read as one value containing an empty one.

**A trailing `.0` is a spreadsheet artifact, never intent.** `3.0` has to reach the
platform as an integer or deserialization into `i64` fails. But `1.0.0` is a
version and `.0` is text, so the test is digits-dot-zeros and nothing looser.

**Only bracketed text is parsed as JSON** — that is how a whole `voting_channels`
array fits in one cell — so a candidate named `NaN` survives, and a description
that opens with a bracket but is not JSON stays a description.

**Naive timestamps are read as UTC.** A cell carries no timezone, and guessing the
author's local one would silently shift a voting window.

**Lists replace rather than concatenate on merge.** Three channels means those
three, not those three appended to the template's.

**Columns that disagree about a field's shape are reported, not panicked.**
`presentation` as a scalar and `presentation.i18n` cannot both hold. New
`Code::ConflictingColumns`, raised while reading a source document rather than
while validating a bundle: nothing can be built until the author picks one, and the
author is who has to see the message.

## Why `Cell` exists

A neutral six-variant vocabulary, so this layer has no opinion about where a row
came from: `.xlsx`, a CSV, or a browser's form state all narrow to the same thing.
The reader that produces `Cell`s is then the only part that needs a spreadsheet
library — which keeps that dependency out of the WASM build for front ends that
have no workbook to read.

## Verified

- `cargo test -p sequent-core --lib --features default_features election_config` — **99 passed**, 0 failed (72 before, 27 new).
- `cargo clippy -p sequent-core --features default_features --lib` — nothing on this module.

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/sequent-core/src/election_config/mod.rs
+++ b/packages/sequent-core/src/election_config/mod.rs
@@ -24,6 +24,7 @@
 //! and <https://github.com/sequentech/meta/issues/12769>.
 
 pub mod emit;
+pub mod paths;
 pub mod problem;
 pub mod report;
 pub mod schema;
@@ -33,6 +34,7 @@ pub mod validate;
 mod validate_tests;
 
 pub use emit::{json_csv, plain_csv, JsonField};
+pub use paths::{coerce_cell, deep_merge, expand, Cell};
 pub use problem::{Code, Problem, Report as ValidationReport, Severity};
 pub use report::{EReportEncryption, Report, ReportCronConfig, ReportType};
 pub use schema::ImportElectionEventSchema;
--- a/packages/sequent-core/src/election_config/paths.rs
+++ b/packages/sequent-core/src/election_config/paths.rs
@@ -1,0 +1,578 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Dotted column headers, cell coercion, and deep merge.
+//!
+//! An authoring spreadsheet's column headers are paths into the target JSON:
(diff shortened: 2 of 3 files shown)


edulix:

PR #2972, opened 2026-08-06
✨ Read an authoring workbook in the shared core

edulix:

**Stacked on https://github.com/sequentech/step/pull/2971.** Base branch is
`feat/meta-12769-paths/main`, so this diff shows only what is added on top.

Sheets of rows of coerced cells, and the `.xlsx` reader that produces them — the
last piece that existed only in janitor's Python, and the one `step-cli` and a
browser both need before either can build a bundle.

## Split in two on purpose

`election_config::sheet` is **pure and knows no file format**. It takes grids of
`Cell` and does the header row, the duplicate-column check, blank-row trimming and
the per-sheet multi-value decision.

`election_config::xlsx` is the **only** part that needs a spreadsheet library, and
does nothing but narrow `calamine`'s cells to `Cell`.

So the awkward cases are all testable without a fixture file. The reader's own
tests build a real `.xlsx` in memory rather than committing a binary — nothing to
keep in step with the code, and no route for a client's data into the repository.

Behind a new **`election_config_xlsx`** feature: admin-portal and voting-portal
enable `default_features` for their WASM build and have no workbook to read, so
they should not carry a spreadsheet parser in their bundle. Reads from bytes rather
than a path, so the same call serves `step-cli` and a browser holding a file input.

## What the tests pinned down

**A duration cell is asked about before being converted.** `calamine`'s
`as_datetime` turns `0.5` into `1899-12-31T12:00` — a plausible-looking timestamp
and entirely wrong — so `is_duration()` is checked first and the number handed over
for validation to object to. Found by a test that assumed the opposite.

**A text cell that looks numeric stays text**, matching the Python. A spreadsheet
hands over a number for a numeric cell, so text here means the author formatted the
column as text — which is how member id `007` stays `007` instead of silently
failing to match a voter. A number typed as a number arrives as a float and does
become an integer.

**Blank rows are dropped rather than trusted.** A spreadsheet's stored dimensions
count rows that were merely visited: the SEIU1000 sample claims a thousand rows for
a sheet holding two. Dropping one must not shift the numbers after it, because the
number is what an author looks at.

**Two tabs that normalise to the same key are refused.** `Admin Users` beside
`AdminUsers` is not a document with a duplicate tab; it is one where nobody knows
which tab is live, and picking silently would eventually import the wrong voters. A
duplicated column is refused for the same reason: which one wins would be arbitrary
and invisible.

**Sheets nobody reads are listed** rather than ignored, so a misspelled tab gets
noticed.

**Absent and empty read the same.** No `Reports` sheet and an empty `Reports` sheet
both mean no reports, so no caller has to special-case either.

**`Origin` carries the sheet name and the row number as the spreadsheet shows
them.** A bundle path like `elections[3].title` is no use to whoever has to fix the
file.

## Known gap

A problem found while reading carries its `Origin` flattened into `Problem::path`.
A front end that wants to highlight the offending cell needs sheet, row and column
separately, which means a structured origin on `Problem` — worth doing when there
is a UI to consume it, not before. Noted in the module docs.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx election_config` — **133 passed**, 0 failed (99 before, 34 new).
- `cargo check -p sequent-core --features default_features` clean — the feature gate holds, `calamine` is not pulled in without it.

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/Cargo.lock
+++ b/packages/Cargo.lock
@@ -632,7 +632,7 @@ dependencies = [
  "attohttpc",
  "home",
  "log",
- "quick-xml",
+ "quick-xml 0.32.0",
  "rust-ini 0.21.3",
  "serde",
  "thiserror 1.0.69",
@@ -1762,6 +1762,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "calamine"
+version = "0.26.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "138646b9af2c5d7f1804ea4bf93afc597737d2bd4f7341d67c48b03316976eb1"
+dependencies = [
+ "byteorder",
+ "chrono",
+ "codepage",
+ "encoding_rs",
+ "log",
+ "quick-xml 0.31.0",
+ "serde",
+ "zip 2.4.2",
+]
+
+[[package]]
 name = "cap-fs-ext"
 version = "3.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


edulix:

PR #2973, opened 2026-08-06
✨ Share the deterministic ids a generated bundle uses

edulix:

**Stacked on https://github.com/sequentech/step/pull/2972.** Base branch is
`feat/meta-12769-workbook/main`, so this diff shows only what is added on top.

Every id in a generated bundle is a version 5 UUID over the event's `external_id`,
the entity kind and the row's `external_id`. Two consequences matter:

- Regenerating an unchanged source produces **byte-identical output**, so a diff
  between two runs shows only what the author actually changed.
- Two events built from different sources **never collide**, because the event's
  `external_id` is mixed into the namespace.

The importer rewrites every UUID it receives, so these are not the ids the platform
ends up storing. They exist to make the file reproducible and its internal
references consistent.

## It agrees with the Python, exactly

`it_agrees_with_the_python_it_replaces` pins six ids produced by janitor's
`ids.py` — namespace, event, election, contest, area/contest link, tenant. They
matched on the first run. That is what lets the Rust take over without renumbering
any event already generated, the SEIU1000 bundle included.

## Written out rather than taken from the `uuid` crate

Two reasons, both deliberate:

- The crate's `v4` feature pulls `getrandom`, whose WASM support is
  version-specific and already pinned elsewhere in this workspace. Nothing here
  needs randomness and it should not acquire a reason to.
- The byte layout is the thing that **must never change** — alter it and every
  event ever generated renumbers. Better read than trusted.

`sha1` is the only new dependency: pure Rust, no rng, and already in
`packages/Cargo.lock`.

## Two details that are easy to get wrong

**Parts are length-prefixed, not joined by a separator.** Joined by one, `["a/b"]`
and `["a", "b"]` hash alike — which for an area/contest link means two different
pairs sharing an id, and one silently overwriting the other. An `external_id`
holding a slash is unusual, not forbidden.

**The prefix counts characters, not bytes**, matching the Python's `len()`. A byte
count is the more obvious choice in Rust and would renumber every id derived from a
non-ASCII `external_id`, for no gain — either count is unambiguous. `José-Muñoz` is
ten characters and twelve bytes, and a test pins it against the Python's answer.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx election_config` — **143 passed**, 0 failed (133 before, 10 new).

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/Cargo.lock
+++ b/packages/Cargo.lock
@@ -9129,6 +9129,7 @@ dependencies = [
  "serde_json",
  "serde_path_to_error",
  "serde_urlencoded",
+ "sha1",
  "sha256",
  "strand",
  "strum 0.27.2",
--- a/packages/sequent-core/Cargo.toml
+++ b/packages/sequent-core/Cargo.toml
@@ -128,6 +128,12 @@ csv = "1.3.0"
 # do not carry it into their WASM bundle
 calamine = { version = "0.26", features = ["dates"], optional = true }
 
+# uuid5 for the deterministic ids a generated bundle uses. The `uuid` crate is not
+# used for this: its v4 feature pulls getrandom, whose WASM support is version
+# specific and already pinned elsewhere in this workspace, and nothing here needs
+# randomness.
+sha1 = { version = "0.10", optional = true }
+
 
 # WASM plugin management
 wasmtime = {version = "41.0.2", optional = true}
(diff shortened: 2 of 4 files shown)


edulix:

PR #2974, opened 2026-08-06
✨ Share the base entity templates and how they render

edulix:

**Stacked on https://github.com/sequentech/step/pull/2973.** Base branch is
`feat/meta-12769-ids/main`, so this diff shows only what is added on top.

The templates supply the platform boilerplate; the source document overrides it.
That split is what keeps client configuration out of the code — a delivery engineer
who needs a different default supplies their own template instead of patching a
tool. Moving the eight `.hbs` files and the renderer into the shared core is what
lets `step-cli` and a browser produce the same bundle from the same workbook.

**Compiled into the binary rather than read from disk.** A browser has no directory
to read, and fetching eight files before rendering anything would be absurd.
Overrides arrive as text from whoever can obtain them: `step-cli` reads a
directory, a SPA takes an upload.

Behind a new **`election_config_templates`** feature, separate from `reports` —
that one also brings a headless browser and three AWS SDKs, and this needs only the
template engine.

## The interesting decision: escaping

Handlebars escapes for HTML by default, which is the **wrong language** here. A
quote would become `&quot;` — valid JSON holding the wrong text. Turning escaping
off instead would let a stray quote break the document.

So the escape function is a **JSON string escape**. Stricter than what the Python
did, and it means a custom template interpolating something undisciplined still
renders parseable JSON. The builtin templates only interpolate ids, timestamps and
enum values and never need it; client text is deep-merged into the *parsed* result
and never passes through the engine at all.

**A helper's output bypasses the escape function.** `{{json}}` relies on that — the
point is to emit a JSON literal, and a quoted object would not be one — while
`{{default}}` has to escape for itself, because it lands inside a string. Both are
tested, since the difference is invisible until a fallback happens to contain a
quote.

## Other decisions worth a look

**An override for a name nobody renders is refused, not ignored.** `elections.hbs`
in a templates directory is a typo, and silently rendering the builtin would leave
its author staring at output that ignores their edit.

**A template that renders invalid JSON quotes the lines around the failure**, with
the offending one marked:

```
error: templates.area: did not render valid JSON: expected value at line 3 column 10
     1 | {
     2 |   "a": 1,
>    3 |   "b": ,
     4 |   "c": 3
```

A bare parse error against a hundred lines of rendered output is not debuggable.

**Strict mode is off on purpose.** A template referring to a field this entity has
no value for should leave it blank for the merge to fill, not stop the build.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx,election_config_templates election_config` — **159 passed**, 0 failed (143 before, 16 new). Every builtin template is asserted to render valid JSON, which is the only thing that would catch a stray comma in a data file.
- `cargo check -p sequent-core --features default_features` clean — the feature gate holds.

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/sequent-core/Cargo.toml
+++ b/packages/sequent-core/Cargo.toml
@@ -167,6 +167,9 @@ default_features = ["dep:strand", "dep:num-bigint", "dep:tempfile", "dep:ammonia
 # default_features because admin-portal and voting-portal have no workbook to
 # read and should not pay for a spreadsheet parser in their bundle.
 election_config_xlsx = ["default_features", "dep:calamine"]
+# Rendering the base entity templates. Separate from `reports`, which also brings
+# a headless browser and three AWS SDKs; this needs only the template engine.
+election_config_templates = ["default_features", "dep:handlebars"]
 
 sqlite = ["dep:tokio", "dep:rusqlite"]
 time = ["dep:time"]
--- a/packages/sequent-core/src/election_config/mod.rs
+++ b/packages/sequent-core/src/election_config/mod.rs
@@ -27,6 +27,12 @@ pub mod emit;
 pub mod ids;
 pub mod paths;
 pub mod problem;
+
+/// Rendering the base entity templates, behind its own feature so a front end
+/// that only validates an existing bundle carries no template engine.
+#[cfg(feature = "election_config_templates")]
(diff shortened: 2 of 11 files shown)


edulix:

PR #2975, opened 2026-08-06
✨ Build the bundle's JSON document in the shared core

edulix:

**Stacked on https://github.com/sequentech/step/pull/2974.** Base branch is
`feat/meta-12769-render/main`, so this diff shows only what is added on top.

Rows to entities: each one a rendered template with the row's dotted-path columns
deep-merged over it, identified by a uuid5, and joined to the others by
`external_id`. Nothing reaches for a clock or a filesystem, so the same build runs
in `step-cli` and in a browser and produces the same bytes.

## The test that makes sharing this worth anything

`what_it_builds_is_a_bundle_the_platform_accepts`: the document the builder
produces **deserializes into `ImportElectionEventSchema`** — the importer's own
struct — and **passes `validate()`** — the importer's own rules. Two
implementations that merely looked similar would not. Before this stack, janitor's
Python could and did produce bundles the platform quietly mangled.

## Problems accumulate, and each names a cell

An author fixing a spreadsheet wants the whole list, not one round trip per
mistake. A bundle path like `contests[3].election_id` is no use to whoever has to
edit the file, so every problem carries the sheet and row as the spreadsheet shows
them:

```
error: sheet 'Contests' row 2 column 'election.external_id': no election has external_id 'nowhere'
```

A test with three separate mistakes asserts all three are reported in one run.

## Decisions carried over, each now with a test saying why

**A control column is consumed, not merged.** `election.external_id` is how a
contest names its election, not a field called `external_id` on an object called
`election`.

**Areas are read in two passes**, so a parent may appear below its own child —
authors do not sort their spreadsheets topologically. An area may not be its own
parent.

**An area needs a name, and two areas may not share one**, because the voters CSV
resolves a voter's area *by name*. A duplicate silently assigns voters to whichever
one the importer finds first.

**A base export is merged under the templates, never over them**, and its identity
fields are scrubbed first: its ids, its bulletin board, its keys, and a
`statistics`/`status` block describing a run that already happened. Carrying any of
those over produces an event that looks configured and is not. Identity is
reasserted *after* the merge, so adding a field cannot reintroduce the bug.

**`reports` stays `[]` and `scheduled_events` stays `null`.** Both travel in their
own CSV; a populated array here is silently dropped, which is how a report goes
missing without an error.

**A parameter nothing interprets is recorded in `election_event.annotations` with a
warning**, not dropped — dropping it is how a setting goes missing on election day.
A parameter with no value is a placeholder the author left blank, and says so.

**An id typed as a number matches the same number written as text.** Whether a cell
was formatted as a number is not something an author controls per column.

## One message improved rather than ported

The Python's `"a election reference is required"` is now phrased without an article
to get wrong: `'election.external_id' is required: it names the election this row
belongs to`.

## What is not here yet

This level builds the JSON document. The CSV members (voters, scheduled events,
reports, admin users, permissions, communication templates), the auth presets and
the Keycloak realm are the next levels of the same port — `keycloak_event_realm` is
`null` until then. Said out loud in the module docs so nobody reads the omission as
a decision.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx,election_config_templates election_config` — **200 passed**, 0 failed (159 before, 41 new).
- `cargo check -p sequent-core --features default_features` clean — the feature gate holds.

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/sequent-core/src/election_config/build.rs
+++ b/packages/sequent-core/src/election_config/build.rs
@@ -1,0 +1,1095 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! Turning a source document's rows into an import bundle.
+//!
+//! Each entity is a rendered template with the row's dotted-path columns
+//! deep-merged over it, identified by a [`super::ids::IdFactory`] uuid5, and
+//! joined to the others by `external_id`. Nothing here reaches for a clock or a
+//! filesystem, so the same build runs in `step-cli` and in a browser and produces
+//! the same bytes.
+//!
+//! Problems accumulate rather than stopping at the first one: an author fixing a
+//! spreadsheet wants the whole list, not one round trip per mistake. Every
+//! problem carries the sheet and row it came from, because a bundle path is no
+//! use to whoever has to edit the file.
+//!
+//! This level builds the JSON document — the event and its elections, contests,
+//! candidates, areas and ballot links. The CSV members (voters, scheduled events,
+//! reports, admin users), the auth presets and the Keycloak realm are the next
+//! levels of the same port; `keycloak_event_realm` is left null until then.
+
+use crate::election_config::ids::IdFactory;
+use crate::election_config::paths::{deep_merge, set_path, split_path};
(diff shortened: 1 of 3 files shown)


edulix:

PR #2976, opened 2026-08-06
✨ Build the bundle's CSV members in the shared core

edulix:

**Stacked on https://github.com/sequentech/step/pull/2975.** Base branch is
`feat/meta-12769-builder/main`, so this diff shows only what is added on top.

Four of a bundle's parts are not in the JSON document at all: voters and scheduled
events are always CSVs, reports are a CSV or nothing, and admin users, role
permissions and communication templates are tenant- or portal-scoped rather than
part of an event import. All of them resolve `external_id`s against the entities the
previous level built, which is why they live beside it — `build_tables.rs` is a
child module of `build.rs` so the two share the resolved ids while each file stays
readable.

**The scheduled-events table is where the voting window actually lives.**
`scheduled_events` in the JSON document is not read by the importer at all. Its rows
go out through `emit`'s `JsonField`, so a SQL NULL is a bare `null` and the payload
is a CSV-quoted JSON literal; the task id comes from the same function that mirrors
`generate_manage_date_task_name`, because a different shape means a task that never
fires.

## What the tests pin down

**A voter's area travels as a name**, because that is what the importer resolves by.

**A voter with no `authorized-election-ids` is authorized for all of them.** Writing
an empty attribute would deny access to every election instead.

**A voter with an email address is treated as verified.** An unverified address
blocks delivery of the one-time code, and a census address is one the client asserts
is correct.

**Any column the builder does not derive becomes a Keycloak user attribute** — how a
client adds a reporting breakout column with no code change.

**A passthrough column blank for every voter is dropped**, and that is not cosmetic:
`get_copy_from_query` treats the mere presence of a `password` header as "hash a
password for each of these voters", so a blank one would give every voter an empty
credential.

**A column name outside `^[a-zA-Z0-9._-]+$` is refused here**, because both CSV
importers reject it mid-import with nothing naming the column.

**An author may write an event type the way they say it** — `start voting period`
and `start-voting-period` both reach `START_VOTING_PERIOD`.

**An election whose window never opens or never closes is warned about.** It imports
fine and then quietly never opens. An event-wide row covers every election.

**A report's `encryption_policy` and `permission_label` are read from the row, not
from the rendered template.** They are control columns, so the row was excluded from
the merge and the template value is only a default — reading it instead is how
`configured_password` silently became `unencrypted` once.

**The permission matrix is transposed.** A matrix is what a human can check at a
glance; `role,permissions` is what `export_tenant_config.rs` writes.

**A template document gets its newlines back.** Literal `\n` and `\"` are what
survive a copy-paste out of a JSON export. Unescaped in one pass, so an escaped
backslash before an `n` does not become a line break.

## One dead branch removed rather than ported

The Python warns when a census has no email or mobile column at all — but `email` is
a *derived* column and is therefore always present, so that branch could never fire.
A census with no contact column reaches the per-voter count with every voter
unreachable, which is the more useful message anyway. A test says so.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx,election_config_templates election_config` — **250 passed**, 0 failed (200 before, 50 new). Includes an end-to-end test that builds every member from one document and asserts the JSON half still deserializes into `ImportElectionEventSchema` and passes `validate()`, plus one that renders the tables through `emit`'s writers.
- `cargo check -p sequent-core --features default_features` clean; `cargo check -p windmill` clean.

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/sequent-core/src/election_config/build.rs
+++ b/packages/sequent-core/src/election_config/build.rs
@@ -15,10 +15,12 @@
 //! problem carries the sheet and row it came from, because a bundle path is no
 //! use to whoever has to edit the file.
 //!
-//! This level builds the JSON document — the event and its elections, contests,
-//! candidates, areas and ballot links. The CSV members (voters, scheduled events,
-//! reports, admin users), the auth presets and the Keycloak realm are the next
-//! levels of the same port; `keycloak_event_realm` is left null until then.
+//! The CSV members and the files that travel beside a bundle are in
+//! `build_tables.rs`, a child module, so that this file stays readable; the two
+//! are one unit of code split across two files.
+//!
+//! The auth presets and the Keycloak realm are the next level of the same port;
+//! `keycloak_event_realm` is left null until then.
 
 use crate::election_config::ids::IdFactory;
 use crate::election_config::paths::{deep_merge, set_path, split_path};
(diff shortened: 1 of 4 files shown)


edulix:

PR #2977, opened 2026-08-06
✨ Configure the login page from the event, in the shared core

edulix:

**Stacked on https://github.com/sequentech/step/pull/2976.** Base branch is
`feat/meta-12769-tables/main`, so this diff shows only what is added on top.

The last of janitor's Python: the four authentication presets, the realm settings an
event already implies, and the permission-label check that explains an empty
Elections list.

## Presets are patches, not realms

A realm is ~165 kB of interdependent Keycloak configuration whose client URLs belong
to the environment it was exported from — and the importer takes
`keycloak_event_realm` **wholesale**, so a present realm *replaces* the
environment's provisioned default rather than merging into it. So a preset is
applied to a realm someone exported from a working event, and is always also kept on
its own so nothing the document asked for is silently dropped. With no base export,
no realm is emitted and the warning says exactly that.

Two things a deep merge cannot express are now **struct fields** rather than magic
keys inside the patch, which is how the Python carried them: binding an
authenticator config, and patching the user profile. That removes the
strip-the-directives step the writer used to need, and with it the chance of
forgetting it.

## The realm work worth reading

**`identityProviders` and `authenticatorConfig` are merged by alias, not replaced.**
They are referenced by alias from elsewhere in the realm, so replacing either
wholesale would strip providers the environment configured on purpose.

**The realm name is derived from tenant and event**, because the voting portal and
the smart-link URLs derive it the same way. A base export's own event id is swapped
out of its client URLs first, because import remaps every UUID it finds and a stale
one would be remapped to something unrelated.

**The user profile is parsed, patched and re-serialised**, not merged: it lives
inside a Keycloak component as a single JSON string.

**A preset naming a flow or authenticator the target realm lacks is warned about**
rather than applied blindly — possible only because each preset states what it needs
and why.

## The event configures its own login page

The platform carries none of this across: it never syncs `supportedLocales`, and has
no path at all from an event name to a realm display name. So the event's own
languages, title and login CSS become realm settings.

Language codes go through the platform's own `iso_639_2t_to_bcp47` rather than a
second copy of it — the Python transcribed all 177 entries by hand. **Basque and
Dutch are missing from that table** and pass through unconverted; both
implementations always behaved that way, so it is a gap in `util::locale` rather
than a regression here, and a test says so.

CSS is escaped for `java.text.MessageFormat` — quotes before braces, so the quotes
this adds are not themselves doubled — and written for **every enabled locale**:
Keycloak looks the message up in the voter's language, so CSS under `en` alone
vanishes the moment a voter switches to Spanish. A stylesheet copied out of a
working realm arrives already quoted and is unwrapped first.

**`uses_otp` now gates the voter-reachability warning.** Under SAML or digital
certificates the client's identity provider authenticates the voter, so "56 of 56
voters cannot be sent a one-time code" is noise rather than a finding.

## The permission-label check

The one that matters most in practice. An election whose label no administrator
holds **imports cleanly, reports no error, and then does not appear in the Elections
list at all**. It happened on the first real import of the SEIU1000 event, where the
document labelled an election `dlc-officers-dburs` while its own administrators
carried `dlc-officers`. Nothing anywhere said so.

Warnings rather than errors, because the document is not the whole picture:
administrators may already exist in the target tenant carrying labels this file
knows nothing about.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx,election_config_templates election_config` — **315 passed**, 0 failed (250 before, 65 new).
- `cargo check -p sequent-core --features default_features` clean; `cargo check -p windmill` clean.
- A test asserts no preset writes an environment-specific host, which is the property that makes them portable at all.

## Screenshots/videos Before

`<empty>`

### DIFF
--- a/packages/sequent-core/src/election_config/branding.rs
+++ b/packages/sequent-core/src/election_config/branding.rs
@@ -1,0 +1,442 @@
+// SPDX-FileCopyrightText: 2026 Sequent Tech Inc <legal@sequentech.io>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+//! What an election event already states about itself, expressed as realm
+//! settings.
+//!
+//! The languages an event enables, its name, and its login CSS all belong on the
+//! login page, and the platform copies none of them there: it syncs the default
+//! locale only under a force-default detection policy, never syncs
+//! `supportedLocales`, and has no path at all from an event name to a realm
+//! display name. Stating them twice in a source document would be a way to get
+//! them out of step, so they are derived from what the event already says.
+//!
+//! The language codes go through [`crate::util::locale::iso_639_2t_to_bcp47`] —
+//! the platform's own table, rather than a second copy of it. The Python this was
+//! ported from transcribed all 177 entries by hand, which is exactly the kind of
+//! duplication this work exists to remove.
+
+use crate::util::locale::iso_639_2t_to_bcp47;
+use serde_json::{json, Map, Value};
+
+/// The realm localization key the login theme renders.
+///
+/// See `sequent-theme/.../sequent.admin-portal/login/template.ftl`.
+pub const LOGIN_CUSTOM_CSS_KEY: &str = "loginCustomCss";
+
+/// Escape CSS for Keycloak's `${msg(...)}`, which is `java.text.MessageFormat`.
+///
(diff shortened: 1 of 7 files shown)


edulix:

PR #2978, opened 2026-08-06
✨ Turn a bundle into the files it ships as

edulix:

**Stacked on https://github.com/sequentech/step/pull/2977.** Base branch is
`feat/meta-12769-realm/main`, so this diff shows only what is added on top.

The last piece before a CLI: what a built bundle is written as, and the zip the
Admin Portal accepts.

Pure, like the rest of the module — named byte blobs, no filesystem. `step-cli`
writes them to a directory, a browser offers them as downloads, and neither leaves a
half-written output behind, because nothing is written until everything is built.

## The line between the two groups

**Importable** members go inside the zip. **Auxiliary** files go beside it:
administrators, roles and communication templates are tenant- or portal-scoped, and
putting them in the zip would mean importing an election event could silently create
administrator accounts. A test asserts `admin_users.csv` is in one list and not the
other.

## Two opposite rules, both deliberate

**The scheduled-events member is written even when empty.** The voting window lives
in it, so whether the file exists must not depend on whether the source had a sheet.

**The reports member is absent when there are none**, because an empty reports CSV is
not a valid one.

## The realm patch says what happened to it

Written whether or not it was applied, with a comment saying which — the two cases
need opposite things done next:

```json
{
  "_comment": "Realm changes this source asks for. NOT applied: no base export was given, …",
  "auth_preset": "voter_link_plus_dob",
  "patch": { … },
  "bind_authenticator_config": {
    "_comment": "Not a merge: point every execution of this authenticator at this config alias.",
    …
  },
  "user_profile": {
    "_comment": "Not a merge: the user profile is a stringified JSON blob inside the … component.",
    …
  }
}
```

Because the two directives are struct fields on `RealmPatch` rather than magic keys
inside the patch, there is nothing to strip out here — so the file can instead state
each one explicitly with a note that it is not a merge. Whoever applies the patch by
hand cannot deduce either from the patch itself.

## The zip is reproducible

A fixed timestamp and a fixed mode on every entry. Without them the archive's bytes
change on every run and "regenerating produced no diff" stops being something anyone
can check. A test zips the same bundle twice and compares; another reads every member
back out and asserts it survived byte for byte.

Behind a new **`election_config_archive`** feature, with the layout half behind the
builder's. A front end that only validates an existing bundle has nothing to write
and should carry no zip writer.

## One deliberate difference from the Python

Every JSON file is two-space indented, rather than one-space for the event document
and two for the rest. Cosmetic — the importer parses it — but it does mean the first
regeneration of an existing event reindents that file once.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx,election_config_archive election_config` — **328 passed**, 0 failed (315 before, 13 new).
- `cargo check -p sequent-core` clean with `default_features` alone, and with `election_config_templates` but not `election_config_archive` — both gates hold.

## Screenshots/videos Before

`<empty>`

_Not applicable — this level is a library module with no UI. The next level, the
`step-cli` subcommand, is where there is output to show._

### DIFF
--- a/packages/sequent-core/Cargo.toml
+++ b/packages/sequent-core/Cargo.toml
@@ -134,6 +134,10 @@ calamine = { version = "0.26", features = ["dates"], optional = true }
 # randomness.
 sha1 = { version = "0.10", optional = true }
 
+# writing the importable archive; optional so a front end that only validates a
+# bundle carries no zip writer
+zip = { version = "2.1", optional = true }
+
 
 # WASM plugin management
 wasmtime = {version = "41.0.2", optional = true}
@@ -142,8 +146,6 @@ wasmtime-wasi = {version = "41.0.2", optional = true}
 [dev-dependencies]
 ptree = "0.5"
 wasm-bindgen-test = "0.3.42"
-# writes the tiny .xlsx the reader's tests read, so no binary fixture is committed
-zip = "2.1"
 
 [features]
 wasmtest = ["wasm", "dep:web-sys", "strand/wasmtest"]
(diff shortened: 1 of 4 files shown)


edulix:

PR #2979, opened 2026-08-06
✨ Fold janitor into step-cli

edulix:

**Stacked on https://github.com/sequentech/step/pull/2978.** Base branch is
`feat/meta-12769-archive/main`, so this diff shows only what is added on top.

janitor, folded into `step-cli`:

```bash
step-cli step build-election-event -w workbook.xlsx -o out --auth-preset none
```

Every decision it makes lives in `sequent_core::election_config` — the same module
windmill validates with and the browser-side tools will run. The command itself only
talks to the filesystem and to whoever ran it. Reading a workbook, building,
validating and laying out the files are all pure functions in the core, which is why
nothing is written until every one of them has succeeded: a workbook with a dangling
reference leaves no half-written output directory behind.

## It produces the same bundle the Python does

Run against the real SEIU1000 workbook and diffed against janitor's output for the
same file, with the same `--tenant-id`:

| | |
|---|---|
| `export_voters-….csv` | **byte-identical** |
| `export_scheduled_events-….csv` | **byte-identical** |
| `export_reports-….csv` | **byte-identical** |
| `export_permissions-….csv` | **byte-identical** |
| `admin_users.csv` | **byte-identical** |
| `templates/*.hbs`, `templates.json` | **byte-identical** |
| `keycloak_admin_realm_patch.json` | **byte-identical** |
| `export_election_event-….json` | **identical when parsed** (2-space indent instead of 1) |
| `keycloak_event_realm_patch.json` | same keys, same patch body |

The event id and the derived tenant id match the Python's to the character, which is
what the pinned `ids` tests predicted. Every warning matches too, including the
`dlc-officers-dburs` permission label that made every election invisible on the first
real import.

## The diff caught a real bug

`admin_users.csv` did *not* match at first: an international phone number arrived as
`33645312453` instead of `+33645312453`.

The workbook computes contact details with a formula, so the cell is `t="str"` with a
cached string result. calamine 0.26 tried a float parse on that string first — and
Rust's float parser accepts a leading `+` — so the number lost its country prefix,
silently. `Utils.sendCode` would then have texted a number with no country code.

Fixed by bumping calamine to **0.36**, where `t="str"` returns a string
unconditionally. Two tests pin it: one for the leading plus, one for a formula result
that is text but looks numeric (a member id of `007`).

This is the class of defect the whole unification exists to catch, and it was caught
by having two implementations to compare rather than by reading the code.

## What the command does

- `--check-only` validates and reports, writing nothing.
- `--strict` refuses to write when there are warnings. Worth having in CI: a warning
  means the bundle imports and the configuration probably is not what its author
  meant.
- `--base-export` takes a `.json` or an export `.zip`, reading
  `export_election_event*.json` out of the archive.
- `--templates-dir` overrides any of the eight entity templates, and says which it
  took. A `.hbs` file whose name is not an entity template is reported rather than
  ignored, because that is a typo and rendering the builtin instead would leave its
  author staring at output that ignores their edit.
- `--auth-preset none` builds without configuring authentication — which the SEIU
  workbook needs, since it declares SAML and leaves the IdP metadata URL blank
  pending the client's identity provider.
- Unread sheets are named, so a misspelled tab does not silently drop its entities.
- The output distinguishes what to import from what travels beside it, and says
  outright that `admin_users.csv` is a secret.

Validation runs twice on purpose, and they are different questions: the builder
reports what is wrong with the *workbook*, in sheet-and-row terms an author can act
on, and `validate()` then reports what would be wrong with the *bundle* — the same
check windmill runs before importing.

## Verified

- `cargo test -p sequent-core --lib --features election_config_xlsx,election_config_archive election_config` — **331 passed**, 0 failed (328 before, 3 new).
- `cargo check` clean for `sequent-core` (each feature gate), `windmill` and `step-cli`.
- No warnings from the new command; `cargo fmt --check` and `cargo clippy` clean.
- End-to-end against the real SEIU1000 workbook, as above. The workbook and its output
  were kept out of the repository and deleted from the build host afterwards.

## Screenshots/videos Before

`<empty>`

```
built seiu1000-leadership-2027
  import this: out/seiu1000-leadership-2027/seiu1000-leadership-2027.zip
    export_election_event-e8e06504-….json
    export_voters-e8e06504-….csv
    export_scheduled_events-e8e06504-….csv
    export_reports-e8e06504-….csv
  beside it, not part of the import:
    out/seiu1000-leadership-2027/admin_users.csv
    out/seiu1000-leadership-2027/export_permissions-….csv
    out/seiu1000-leadership-2027/templates/…
    out/seiu1000-leadership-2027/keycloak_event_realm_patch.json
  warning: admin_users.csv may carry clear-text passwords. It is a secret, not a deliverable.
```

_A screenshot of the imported event follows once this is run against a live
environment._

### DIFF
--- a/packages/Cargo.lock
+++ b/packages/Cargo.lock
@@ -527,6 +527,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "atoi_simd"
+version = "0.18.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f3cdb3708a128e559a30fb830e8a77a5022ee6902806925c216658652b452a44"
+dependencies = [
+ "debug_unsafe",
+ "rustversion",
+]
+
+[[package]]
 name = "atomic"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1763,18 +1773,20 @@ dependencies = [
 
 [[package]]
 name = "calamine"
-version = "0.26.1"
+version = "0.36.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "138646b9af2c5d7f1804ea4bf93afc597737d2bd4f7341d67c48b03316976eb1"
+checksum = "5fa68281b1a76b54a62156474adb06bb380a67e07dd60656e3217152b42183f3"
 dependencies = [
+ "atoi_simd",
  "byteorder",
  "chrono",
  "codepage",
  "encoding_rs",
+ "fast-float2",
  "log",
- "quick-xml 0.31.0",
+ "quick-xml 0.41.0",
  "serde",
- "zip 2.4.2",
+ "zip 8.6.0",
 ]
 
 [[package]]
@@ -3043,6 +3055,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "debug_unsafe"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7eed2c4702fa172d1ce21078faa7c5203e69f5394d48cc436d25928394a867a2"
+
+[[package]]
 name = "debugid"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


edulix:
Superseded by https://github.com/sequentech/step/pull/2983. Fifteen stacked PRs was a review burden rather than a service, so the same commits are now three: [2981](https://github.com/sequentech/step/pull/2981) the shared definition and windmill using it, [2982](https://github.com/sequentech/step/pull/2982) building a bundle from a workbook, [2983](https://github.com/sequentech/step/pull/2983) `step-cli` and the browser over it. Nothing was squashed — every commit and its reasoning is still there, just grouped.

PR #2980, opened 2026-08-06
✨ Give the browser the same checks the server runs

edulix:

**Stacked on https://github.com/sequentech/step/pull/2979.** Base branch is
`feat/meta-12769-cli/main`, so this diff shows only what is added on top.

`election_config::wasm` is a thin wrapper and nothing more: the work happens in the
same functions `step-cli` and windmill call, and this converts between them and
JavaScript. That is the point — **a file that validates in a browser imports on the
server, because the same code decided both times.**

## `checkBundle` ships today, with no workflow change

`election_config::wasm` is gated the way `crate::wasm` is, and `build_wasm.yml`
already builds with `wasmtest,default_features` — which enables both. So the admin
portal can tell an operator what is wrong with an export *before* they upload it, and
the answer is the importer's own.

```ts
import {checkBundle} from "sequent-core";

const report = checkBundle(await file.text());
report.problems.forEach(p => console.log(p.severity, p.code, p.path, p.message));
```

`buildFromWorkbook` and `authPresets` are **additionally** gated on the xlsx and
archive features, which that build does not enable. So admin-portal, voting-portal,
ui-core and ballot-verifier gain the validator and carry no spreadsheet parser, no
template engine and no zip writer. Packaging a build that *does* enable them lands
with the SPA that needs it, rather than being guessed at here.

## The finding: `zip`'s defaults cannot compile for a browser

The `zip` crate's default features pull **bzip2, zstd and lzma** — C libraries with
no wasm32 target. Left as they were, nothing in this module could ever have compiled
for a browser, which is the whole point of the work, and the failure would have
surfaced much later in a front end's build:

```
error: failed to run custom build command for `bzip2-sys v0.1.13+1.0.8`
error: failed to run custom build command for `zstd-sys v2.0.16+zstd.1.5.7`
error: failed to run custom build command for `lzma-sys v0.1.20`
```

Now `default-features = false` with `deflate` only, which is the only method an
import archive uses. `calamine` already restricted its own `zip` the same way, which
is how it was obvious this was ours.

## Interface decisions

**Plain JS values, not wasm-bindgen classes.** A front end holds these in state,
hands them to React and serialises them; an opaque handle with a `free()` method is a
memory leak waiting for whoever forgets to call it.

**A failed build returns its problems rather than throwing.** A list of problems is
something a page can render; an exception is not. The one exception is a file that
does not parse as an export at all, where there is no list to render.

**The TypeScript for `Problem`, `Report` and `BuildOutput` is declared here**, not
written again in the front end — and `authPresets()` returns the presets rather than
a dropdown duplicating them, so a UI cannot offer a preset that does not exist or
miss one that does.

## Verified

- `cargo check -p sequent-core --features wasmtest,default_features` and the same plus all three config features — **0 errors each**, identical to the baseline before this change.
- `cargo test … election_config` — **331 passed**, unchanged.
- `cargo check -p windmill` and `-p step-cli` clean; `cargo fmt --check` and `cargo clippy` clean.

**Not verified locally:** the wasm32 cross-compile itself. `ring`'s build script needs
a clang that targets wasm32, which Apple's does not have — the baseline
`wasmtest,default_features` build fails identically on this machine, before and after
this change, so the failure is environmental. CI builds that target today, and the
`zip` fix above is precisely what its first run of the config features would
otherwise have hit.

## Screenshots/videos Before

`<empty>`

_Follows with the SPA that consumes this._

### DIFF
--- a/packages/sequent-core/Cargo.toml
+++ b/packages/sequent-core/Cargo.toml
@@ -134,9 +134,13 @@ calamine = { version = "0.36", features = ["dates"], optional = true }
 # randomness.
 sha1 = { version = "0.10", optional = true }
 
-# writing the importable archive; optional so a front end that only validates a
-# bundle carries no zip writer
-zip = { version = "2.1", optional = true }
+# Writing the importable archive; optional so a front end that only validates a
+# bundle carries no zip writer.
+#
+# default-features = false is load-bearing: the default set pulls bzip2, zstd and
+# lzma, all of which are C libraries with no wasm32 target, and this module has to
+# compile for a browser. Deflate is the only method an import archive uses.
+zip = { version = "2.1", default-features = false, features = ["deflate"], optional = true }
 
 
 # WASM plugin management
--- a/packages/sequent-core/src/election_config/mod.rs
+++ b/packages/sequent-core/src/election_config/mod.rs
@@ -50,6 +50,16 @@ pub mod schema;
 pub mod sheet;
 pub mod validate;
 
+/// The browser's view of this module: thin wrappers over the same functions
+/// `step-cli` and windmill call.
+///
+/// Gated the way `crate::wasm` is, so it appears in exactly the builds that have a
+/// JavaScript side. What it exposes depends on the other features — checking an
+/// existing bundle needs neither a template engine nor a spreadsheet parser, and a
+/// front end that only does that should carry neither.
+#[cfg(all(feature = "wasm", feature = "default_features"))]
+pub mod wasm;
+
 /// Reading `.xlsx`, behind its own feature so front ends with no workbook to read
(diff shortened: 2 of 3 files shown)


edulix:
Superseded by https://github.com/sequentech/step/pull/2983. Fifteen stacked PRs was a review burden rather than a service, so the same commits are now three: [2981](https://github.com/sequentech/step/pull/2981) the shared definition and windmill using it, [2982](https://github.com/sequentech/step/pull/2982) building a bundle from a workbook, [2983](https://github.com/sequentech/step/pull/2983) `step-cli` and the browser over it. Nothing was squashed — every commit and its reasoning is still there, just grouped.

PR #2982, opened 2026-08-07
✨ Build an importable election event from a workbook, in the shared core

edulix:

**2 of 3.** Stacked on https://github.com/sequentech/step/pull/2981, so this diff
shows only what is added on top. [The tools over it](https://github.com/sequentech/step/pull/2983)
follow.

The whole of janitor's Python, ported into `sequent-core::election_config`. Every
module is new and behind a feature, so nothing that does not want it pays for it —
with one deliberate exception, below: windmill's exporter now writes its two CSVs
through the shared code instead of its own.

## The test that makes this worth doing

`what_it_builds_is_a_bundle_the_platform_accepts`: the document the builder produces
deserializes into `ImportElectionEventSchema` — the importer's own struct — and passes
`validate()`, the importer's own rules. Two implementations that merely looked similar
would not.

## What is in it

| Module | Feature | |
|---|---|---|
| `emit` | — | The CSV byte shapes, written once instead of three times |
| `paths` | — | Dotted headers, cell coercion, deep merge |
| `sheet` | — | Table shaping over a neutral `Cell` |
| `ids` | — | The deterministic uuid5 factory |
| `branding` | — | An event's languages, title and login CSS as realm settings |
| `presets` | — | The four authentication presets |
| `xlsx` | `election_config_xlsx` | `calamine`; the only part that knows a file format |
| `render` | `election_config_templates` | The eight `.hbs` entity templates, compiled in |
| `build` | `election_config_templates` | Rows to a bundle |
| `archive` | `election_config_archive` | The files a bundle ships as, and a reproducible zip |

The gates are the point: admin-portal, voting-portal, ui-core and ballot-verifier
build with `default_features` and get the schema and validator, carrying no
spreadsheet parser, template engine or zip writer.

## The decisions worth a reviewer's time

**A blank cell and a cell holding `null` are different things**, all the way through.
Blank means the author said nothing, so the template default holds; `null` means they
cleared it. Collapse them and there is no way to remove a default.

**`emit`'s separator is `|` and `paths`' is `||`.** Two constants because they are two
formats — the workbook's and the importer's — and something has to convert. A `||`
that reaches a CSV is read as one value containing an empty one.

**Ids are length-prefixed, not separator-joined.** `["a/b"]` and `["a", "b"]` would
otherwise hash alike, which for an area/contest link means two pairs sharing an id and
one silently overwriting the other. The prefix counts *characters*, matching the
Python, so no already-generated event renumbers; six ids are pinned against `ids.py`
and matched on the first run.

**Handlebars escapes for JSON, not HTML.** HTML escaping would turn a quote into
`&quot;` — valid JSON holding the wrong text — and no escaping would let a stray quote
break the document. A helper's output bypasses the escape function, which `{{json}}`
relies on and `{{default}}` has to compensate for.

**A base export is merged *under* the templates and scrubbed first** of its ids, its
bulletin board, its keys and a `statistics`/`status` block describing a run that
already happened. Identity is reasserted after the merge, so adding a field cannot
reintroduce the bug.

**`reports` stays `[]` and `scheduled_events` stays `null`** in the JSON document.
Both travel in their own CSV; a populated array here is silently dropped, which is how
a report goes missing without an error.

**A voter's area travels as a name**, because that is what the importer resolves by —
so two areas may not share one.

**A passthrough voter column blank for every voter is dropped.** Not cosmetic:
`get_copy_from_query` treats the mere presence of a `password` header as "hash a
password for each of these voters".

**A report's `encryption_policy` is read from the row, not the template.** It is a
control column, so the row was excluded from the merge and the template value is only
a default — reading it instead is how `configured_password` silently became
`unencrypted` once.

**Presets are patches, not realms.** The importer takes `keycloak_event_realm`
wholesale, so a present realm *replaces* the environment's provisioned default.
`identityProviders` and `authenticatorConfig` merge by alias rather than being
replaced, because other parts of the realm reference them by alias.

**Auxiliary files go beside the zip, never inside it.** Putting administrators in the
archive would mean importing an election event could silently create administrator
accounts.

**The permission-label check.** An election whose label no administrator holds imports
cleanly, reports no error, and then does not appear in the Elections list at all — as
happened on the first real import, where a document labelled an election
`dlc-officers-dburs` while its administrators carried `dlc-officers`.

## windmill's duplicate writers are gone, and one was a trap

The exporter wrote both CSVs itself. It now writes them through `emit`, so an export
and a generated bundle are one shape rather than two that happen to agree.

The scheduled-events writer is worth reading. It derived both the header and each row
from `serde_json::to_value(event).as_object()`, taking `.keys()` and `.values()`. That
produces the right file **today**, but only because three unstated things line up:

1. `preserve_order` is enabled somewhere in the dependency graph, so a
   `serde_json::Map` iterates in insertion order rather than alphabetically;
2. insertion order is `ScheduledEvent`'s field order;
3. that order happens to match what the importer reads.

`import_scheduled_events.rs` takes the payload from `record.get(10)`. Sorted
alphabetically, index 10 is `task_id` and the payload is at 5 — so **every exported
event would import with its payload read as a task name**, and the task that opens
voting would never fire. Reordering the struct, a refactor with no visible risk, or
losing `preserve_order` from the graph, would have done that silently, on the disaster
recovery path.

The row is now built field by field against `SCHEDULED_EVENT_COLUMNS`. The order is
the code rather than a consequence of it.

The reports CSV wrote `"ID"`, `"Election ID"`, … as its header — harmless, since
`process_reports_file` skips it and reads by index, but it meant an export and a
generated bundle differed on sight for no reason.

One byte-level change: `emit` writes `\n` line endings where `csv::Writer` wrote
`\r\n`. The reader accepts either, and it makes an export diffable against a
generated bundle.

## A fixture suite both callers run

The point of one validator is that two callers reach the same answer. Tests written
separately in Rust and TypeScript would prove only that each side agrees with itself.
So the cases are **data** — `fixtures/sound.json` plus `fixtures/cases.json`, compiled
in and handed to the browser through `fixtureCases()` in PR 2983.

A case is a patch, not a whole bundle: one bundle to keep valid, and reading a case
means reading the difference. Eight of them, each carrying the reason it exists,
covering a dangling reference, a duplicate id, an area loop, more winners than
candidates, a tally mismatch, an empty ballot, and an election hidden by a permission
label — the one that cost a real import.

Three tests guard the suite rather than the code: at least eight cases loaded (a loop
over an empty list passes while checking nothing), the base bundle is itself sound,
and every code validation can produce has a case — with the three that deliberately do
not named in the assertion, so adding a code without a case fails rather than passing
quietly.

## The ways of voting, and why they are validated in pairs

`voting_channels` was in `election_event.hbs` and `election.hbs` and settable from
nothing. Four flags, and every one of them is a **precondition rather than a
switch**: turning it on makes the platform *offer* the channel, and something else
has to exist before a voter can use it. That asymmetry is the whole reason this is
validated rather than just written — a flag on with its other half missing produces
a start control in the Admin Portal that opens a period nobody can vote in, found
out on election day by somebody who cannot fix it.

| Channel | Its other half | Verdict |
|---|---|---|
| `online` | nothing | Off is an **error** — nobody could vote |
| `early_voting` | an area allowing it | **Error** either way round, and the message names the areas |
| `kiosk` | an auth client named `<client>-kiosk` | **Warning** — the half is in the environment |
| `telephone` | the event's IVR tab | **Warning** — the half is filled after import |

Early voting's pair is the platform's, not an invention: `ElectionSelectionScreen`'s
`isEarlyVotingOpen` is `area_presentation.allow_early_voting == allow_early_voting
&& early_voting_status == OPEN`, and `Area/FormContent` hides the area's own field
while the event's channel is off — so a bundle with only one half cannot have been
made in the Admin Portal, and the area's setting does nothing.

Written to the event **and every election** from one field, because Publish reads
the block off `useRecordContext<ElectionEvent | Election>`. A difference between
the two levels is now reported, compared by whether each channel is *on* rather
than which keys are present: `election_event.hbs` names three channels and
`election.hbs` names four, and comparing presence would flag every ordinary build
about `telephone`. There is no per-election channel editor anywhere in the
platform, so a bundle in that state was written by hand.

**Two fields deliberately not exposed**, each recorded rather than guessed:

- **`paper`** is in `hasura_core::VotingChannels` and in nothing else — no
  `VotingStatusChannel` variant, no status block, no Publish control, no label, and
  not in the Admin Portal's own `defaultChannels`. A bundle naming it gets a
  warning rather than a refusal: it does no harm, it just does nothing, and
  somebody who set it believes otherwise.
- **`is_kiosk`** on the election stays as the template's `false`. Its only reader
  is `GetTallyData`, which selects the whole row — no behaviour hangs off it, and
  the voting portal decides kiosk-ness from a `?kiosk` URL. Setting it alongside
  `voting_channels.kiosk` would be a guess about an election that allows both.

And a gap closed on the way past: **nothing anywhere refused an invalid
`allow_early_voting`.** An existing builder test overrides `area.hbs` with
`early_voting_allowed` — a plausible rearrangement of the real string. The portal
compares it exactly, so a near miss reads as "no early voting" in silence.

## An image does travel inside a bundle — and I said the opposite first

Worth reading as a correction, because the mistake has a shape worth avoiding.

I concluded a candidate photograph could not survive an import, from two true facts:
`ImportElectionEventSchema` has no `documents` array, and `replace_uuids` rewrites
every identifier in the JSON. Both hold. Neither is the whole picture, because the
bytes travel as a **zip entry** rather than in the JSON — and I never looked at what
the importer does with zip entries. I even noted that `replace_ids` returns a
replacement map, and did not ask who consumes it.

`process_s3_file` consumes it. The `images/` branch of `import_election_event` hands
it that same map, and it

1. pulls the old identifier from the entry name with `document_([0-9a-f-]{36})`,
2. looks it up in the map,
3. creates the document **with the new identifier**, `is_public: true`.

So `image_document_id`, the identifier inside `presentation.urls`, and the uploaded
file all move to the same new value together.

The contract is three parts that must agree:

| | |
|---|---|
| `candidates[].image_document_id` | `<document>` |
| `candidates[].presentation.urls[]` | `{"url": "tenant-<t>/document-<d>/<file>", "is_image": true}` |
| a zip entry | `images/document_<d>_<file>` |

`check_images` validates the half a JSON-only pass can see. **A url naming a
different identifier than the field beside it is an error** — the two are kept
together only because they are the same string before the rewrite, so differing means
they are rewritten to two different values. Each half present without the other is a
warning: a url alone still puts the picture on the ballot but leaves the Admin Portal
unable to change it; a document alone is uploaded and shown nowhere.

Two things left to `engineering/how-an-image-travels-in-a-bundle` because this pass
sees the JSON and not the zip: an entry whose identifier appears nowhere in the JSON
**fails the whole import** (`replacement_map.get(uuid).ok_or_else(…)`, and the map
only holds identifiers that were actually rewritten), and the generator writes no
tempfile prefix — the platform's exports carry one, `extract_document_uuid` matches
unanchored, and omitting it keeps generation deterministic.

**The narrower rule.** What `keys_ceremonies[].trustee_ids` establishes is that an
identifier pointing at a row the bundle does not contain is destroyed. It does *not*
follow that such a thing cannot be carried — the archive can contain the row. The
question is not "is it an identifier?" but "does the archive carry what it points at,
and does the importer know to look?"

No control yet. The tenant is resolved inside the builder — `resolve_tenant_id` draws
on the explicit option, the Parameters sheet, the base export and the id-factory
fallback — so only the builder can compose the url, and the plan and UI half is its
own change.

## Things fixed rather than ported

- **A warning branch the Python could never reach**: it warns when a census has no
  email or mobile column, but `email` is a *derived* column and is always present.
- **`iso_639_2t_to_bcp47` is the platform's own table**, not a second hand
  transcription of all 177 entries.
- **`"a election reference is required"`** is reworded to have no article to get wrong.

## Verified

- `cargo test … election_config` — **536 passed**, 0 failed.
- `cargo check -p sequent-core` clean at each feature gate, and `-p windmill`,
  `-p harvest` and `-p step-cli` clean individually.

  This claim was **false when first written**, and said so for long enough to
  redden fourteen jobs. `pub mod build` is gated on `election_config_templates`
  and `pub use build::{…}` was not, so the crate did not compile for any feature
  set that leaves the templates out — which is every consumer except step-cli.
  It was clean at the gate I checked, not at the default one, and
  `cargo build --workspace` hides it by unifying features across the graph.
  Fixed in `🐞 Gate the builder's re-export like the module it names`, with six
  `cargo check --lib` jobs added to `tests.yml` so the next one fails in two
  minutes and names the gate.
- `cargo build` emits no warnings for these modules; `cargo fmt --check` and `cargo clippy --lib --tests` clean.

## Reading it

Eleven commits, each one module and its reasoning: `emit`, `paths`, `sheet`+`xlsx`,
`ids`, `render`, `build`, the CSV members, presets and the realm, `archive`, then
windmill's duplicate writers deleted and the fixture suite added.

## Screenshots/videos Before

`<empty>`

_Not applicable — library modules with no UI. See PR 3 for the command's output._

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,6 +10,63 @@ on:
     branches: [ 'main', 'release/[0-9]+.[0-9]+' ]
 
 jobs:
+  # sequent-core is compiled against a different feature set by nearly every
+  # consumer, and a module gated one way with its re-export gated another only
+  # fails for the sets nobody happens to build locally: `cargo build --workspace`
+  # unifies features across the graph and hides it entirely. That is exactly how
+  # an ungated `pub use build::{…}` reached CI and reddened fourteen jobs in the
+  # slow suites below, each of which reports it as its own failure.
+  #
+  # `--lib` rather than `--all-targets`: the question is whether the library
+  # compiles for a consumer that asked for this much and no more. The test
+  # modules need `default_features` and are covered by the suites below.
+  #
+  # This job is `cargo check` only, so it finishes in a couple of minutes and
+  # names the gate that broke.
+  feature-gates:
+    name: sequent-core at ${{ matrix.gate }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - gate: no features
+            cache: none
+            features: ""
+          - gate: default_features
+            cache: default
+            features: --features default_features
+          - gate: keycloak,default_features
+            cache: keycloak
+            features: --features keycloak,default_features
+          - gate: election_config_templates
+            cache: templates
+            features: --features election_config_templates
+          - gate: election_config_archive
+            cache: archive
+            features: --features election_config_archive
+          - gate: election_config_xlsx
+            cache: xlsx
+            features: --features election_config_xlsx
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Set up Rust tests
+        uses: ./.github/actions/setup-rust-tests
+        with:
+          # Per gate, or six jobs would evict each other's artifacts all day.
+          cargo-build-name: sequent-core-${{ matrix.cache }}
+          cargo-build-path: packages/target
+          cargo-lock-path: packages/Cargo.lock
+
+      - name: cargo check ${{ matrix.gate }}
+        env:
+          PROTOC: "/usr/bin/protoc"
+        run: cargo check -p sequent-core --lib ${{ matrix.features }}
+        working-directory: packages
+
   run-tests:
     name: Run Rust tests
     runs-on: ubuntu-24.04
--- a/packages/Cargo.lock
+++ b/packages/Cargo.lock
@@ -9129,6 +9129,7 @@ dependencies = [
  "serde_json",
  "serde_path_to_error",
  "serde_urlencoded",
+ "sha1",
  "sha256",
  "strand",
  "strum 0.27.2",
--- a/packages/Cargo.lock
+++ b/packages/Cargo.lock
@@ -632,7 +632,7 @@ dependencies = [
  "attohttpc",
  "home",
  "log",
- "quick-xml",
+ "quick-xml 0.32.0",
  "rust-ini 0.21.3",
  "serde",
  "thiserror 1.0.69",
@@ -1762,6 +1762,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "calamine"
+version = "0.26.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "138646b9af2c5d7f1804ea4bf93afc597737d2bd4f7341d67c48b03316976eb1"
+dependencies = [
+ "byteorder",
+ "chrono",
+ "codepage",
+ "encoding_rs",
+ "log",
+ "quick-xml 0.31.0",
+ "serde",
+ "zip 2.4.2",
+]
+
+[[package]]
 name = "cap-fs-ext"
 version = "3.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2120,6 +2136,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fa961b519f0b462e3a3b4a34b64d119eeaca1d59af726fe450bbba07a9fc0a1"
 dependencies = [
  "thiserror 2.0.17",
+]
+
+[[package]]
+name = "codepage"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "48f68d061bc2828ae826206326e61251aca94c1e4a5305cf52d9138639c918b4"
+dependencies = [
+ "encoding_rs",
 ]
 
 [[package]]
@@ -7724,6 +7749,16 @@ checksum = "a993555f31e5a609f617c12db6250dedcac1b0a85076912c436e6fc9b2c8e6a3"
 
 [[package]]
 name = "quick-xml"
+version = "0.31.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1004a344b30a54e2ee58d66a71b32d2db2feb0a31f9a2d302bf0536f15de2a33"
+dependencies = [
+ "encoding_rs",
+ "memchr",
+]
+
+[[package]]
+name = "quick-xml"
 version = "0.32.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1d3a6e5838b60e0e8fa7a43f22ade549a37d61f8bdbe636d0d7816191de969c2"
@@ -8535,7 +8570,7 @@ dependencies = [
  "minidom",
  "native-tls",
  "percent-encoding",
- "quick-xml",
+ "quick-xml 0.32.0",
  "serde",
  "serde_derive",
  "serde_json",
@@ -9057,6 +9092,7 @@ dependencies = [
  "aws-smithy-types",
  "base64 0.22.1",
  "borsh",
+ "calamine",
  "cfg-if",
  "chrono",
  "console_error_panic_hook",
@@ -9113,6 +9149,7 @@ dependencies = [
  "wasmtime",
  "wasmtime-wasi",
  "web-sys",
+ "zip 2.4.2",
 ]
 
 [[package]]
--- a/packages/sequent-core/Cargo.toml
+++ b/packages/sequent-core/Cargo.toml
@@ -124,6 +124,10 @@ kuchiki = { version = "0.8", optional = true }
 rusqlite = { version = "0.32", features = ["bundled"], optional = true }
 csv = "1.3.0"
 
+# reading authoring spreadsheets; optional so front ends with no workbook to read
+# do not carry it into their WASM bundle
+calamine = { version = "0.26", features = ["dates"], optional = true }
+
 
 # WASM plugin management
 wasmtime = {version = "41.0.2", optional = true}
@@ -132,6 +136,8 @@ wasmtime-wasi = {version = "41.0.2", optional = true}
 [dev-dependencies]
 ptree = "0.5"
 wasm-bindgen-test = "0.3.42"
+# writes the tiny .xlsx the reader's tests read, so no binary fixture is committed
+zip = "2.1"
 
 [features]
 wasmtest = ["wasm", "dep:web-sys", "strand/wasmtest"]
@@ -151,6 +157,10 @@ lambda_openwhisk = []
 lambda_aws_lambda = []
 plugins_wit = ["dep:wasmtime", "dep:wasmtime-wasi"]
 default_features = ["dep:strand", "dep:num-bigint", "dep:tempfile", "dep:ammonia"]
+# Reading .xlsx into an election_config::sheet::Workbook. Separate from
+# default_features because admin-portal and voting-portal have no workbook to
+# read and should not pay for a spreadsheet parser in their bundle.
+election_config_xlsx = ["default_features", "dep:calamine"]
 
 sqlite = ["dep:tokio", "dep:rusqlite"]
 time = ["dep:time"]
--- a/packages/sequent-core/Cargo.toml
+++ b/packages/sequent-core/Cargo.toml
@@ -134,6 +134,10 @@ calamine = { version = "0.26", features = ["dates"], optional = true }
 # randomness.
 sha1 = { version = "0.10", optional = true }
 
+# writing the importable archive; optional so a front end that only validates a
+# bundle carries no zip writer
+zip = { version = "2.1", optional = true }
+
 
 # WASM plugin management
 wasmtime = {version = "41.0.2", optional = true}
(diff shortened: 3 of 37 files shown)


edulix:
@coderabbitai review

edulix:
Worked through the review — 13 of the 17 addressed across four commits (`ab5a5c8`, `96964d9`, `3ba7736`, `8452d2a`). Each finding was checked against the code first; every one below was real.

**Defects**

- **`build_realm.rs` panicked on a malformed base export.** `.expect("a realm component is an object")` on an array element that can be any JSON value. Reported through `self.warn` now, like the missing-attributes case two blocks above.
- **`build_realm.rs` returned a realm of the wrong type.** A failed reparse after the id swap returned `Value::String(swapped)`, so `keycloak_event_realm` held text where the importer expects an object — and it takes it wholesale. Carries the realm over unchanged and says so.
- **`time.rs` could overflow.** `offset_minutes * 60` is `i32` and `instant()` runs before `check()`; `checked_mul` folds it into the existing problem. Test added.
- **Two schedule rows for one processor and election became one.** Same uuid5, same task id, importer keeps one, other time lost silently. `DuplicateId` naming the earlier row, like `require_unique` does for voters. Test added.
- **A whitespace-only `authorized-election-ids` denied every election.** Present-but-blank fell through the `is_empty` guard to an empty attribute. A blank cell now means what an absent one means.
- **`row.text` read ids differently from `resolve`.** No numeric coercion, no trim — so a numeric `external_id` registered in the first Areas pass and vanished in the second, and in AreaContests two different numeric pairs both keyed the duplicate check as `("", "")`. Both reads go through `value_as_text`. The numeric-id test is extended to both sheets, which is your `build_tests.rs` finding.
- **Plan validation caught only self-parenting.** `climbs_into_a_loop` walks the chain with a visited set, so A→B→A is reported in the plan's own vocabulary. Test added.

**One source of truth**

- **`EVENT_PROCESSORS` is gone.** `event_processor` parses through `EventProcessors::from_str` and returns the typed value; `check_voting_windows` compares variants. The fuzzy normalisation happens before the parse, as before.
- **`scheduled_event_task_id` is gone.** Its own doc said it mirrored `generate_manage_date_task_name` "reproduced rather than approximated" — it calls that function now, and the two tests that asserted the exact task name still do, so they pin the scheduler's shape rather than a copy's. `EventProcessors` gained `EnumIter` so the "expected one of" message names the variants.

**Stale comments**: `names()` returns declaration order, not sorted; `scheduled_event.hbs` pointed at `build.py` for values `election_config::build_tables` computes.

**The fixture coverage check could not fail.** Its comment claimed a new `Code` without a case fails the test; it walked a hard-coded list. Every variant is walked and placed by an exhaustive `match` now — which is how `ConflictingColumns`, missing from the old list, got placed.

**Still open, deliberately:**

- **`contest.hbs` placeholders** — I have not removed them. You are right that the template fills fields before validation sees them, so `MissingField` can never fire for those five. But these templates were ported from janitor and #2983 claims byte-for-byte parity with its output; removing keys makes the builder stricter than the tool it replaces and would reject workbooks that build today. That is a product call about which columns become mandatory, not a code cleanup, so it needs the ticket owner rather than me.
- **`presets.rs` typed `Requirement.kind`** and **`tests.yml` least-privilege `permissions`** — both agreed, not yet done.

577 sequent-core tests, all seven feature gates and `cargo fmt --check` clean, verified in the devcontainer.

edulix:
All seventeen review threads are now answered inline and resolved. The four that were still open at the last pass:

**`presets.rs` typed `Requirement.kind`** — done in `d398529`. `RequirementKind` over `Flow`, `Authenticator` and `AuthenticatorConfig`, with `Display` and `FromStr` over the same lowercase strings. The realm builder matches exhaustively, so a typo is a compile error rather than a silent trip down the authenticator path.

**`tests.yml` least privilege** — done in `d398529`. `permissions: contents: read` on the gate job, `persist-credentials: false` on the checkout.

**A plan may not reuse an identifier** — `2d5e16d`, and it was the one finding nobody had touched. `uid` keys on the kind and the `external_id` and nothing else, so the same identifier twice inside a kind mints one id for two things and the second replaces the first wherever it is referenced. The plan validator caught a duplicate area *name*; the builder caught the identifier, but as a workbook row number, which is not a thing a wizard author ever saw. One helper, `require_unique_external_id`, now used for areas, elections, contests and candidates — the same hole was in all four, and contests and candidates are keyed across the whole plan rather than per election because that is how `uid` keys them. A blank identifier stays with the `MissingField` rule that owns it; two things to fill in are not two names for one thing.

**`contest.hbs` placeholders** — `75b7f11`, addressed but not as prescribed. Removing the five values breaks two callers: `contests_sheet` did not write `voting_type` or `counting_algorithm` at all, so every Architect plan would stop building, and janitor's workbooks — which #2983 matches byte for byte — leave the rest out too. Which columns become mandatory is a product call about the import format. What is not a product call is being *told*: `build_contests` warns per column, naming the value that stood in, and `contests_sheet` now writes those two itself, since the wizard offers one voting method today and a workbook that says which one is a document somebody can read. Same values, so the built bundle is byte-for-byte what it was. The template's comment claimed validation rejects a contest missing any of the five; it does not, and now says so.

583 sequent-core tests, all six feature gates and `cargo fmt --check` clean in the devcontainer.

PR #2983, opened 2026-08-07
✨ step-cli and the browser over the shared core

edulix:

**3 of 3.** Stacked on https://github.com/sequentech/step/pull/2982, so this diff
shows only what is added on top.

Two front ends over the shared core: a `step-cli` subcommand, and the browser's view
of the same functions. Neither adds a decision of its own — that is the point.

```bash
step-cli step build-election-event -w "Client Import Workbook.xlsx" --tenant-id …
```

```ts
import {checkBundle} from "sequent-core";
const report = checkBundle(await file.text());
```

## It produces the same bundle the Python does

Run against the real SEIU1000 workbook and diffed against janitor's output for the
same file, with the same `--tenant-id`:

| | |
|---|---|
| `export_voters-….csv` | **byte-identical** |
| `export_scheduled_events-….csv` | **byte-identical** |
| `export_reports-….csv` | **byte-identical** |
| `export_permissions-….csv` | **byte-identical** |
| `admin_users.csv` | **byte-identical** |
| `templates/*.hbs`, `templates.json` | **byte-identical** |
| `keycloak_admin_realm_patch.json` | **byte-identical** |
| `export_election_event-….json` | **identical when parsed** (2-space indent instead of 1) |

The event id and the derived tenant id match to the character. Every warning matches
too, including the `dlc-officers-dburs` permission label that made every election
invisible on the first real import.

## The diff caught two real bugs

**A phone number lost its `+`.** `admin_users.csv` did not match at first:
`+33645312453` arrived as `33645312453`. The workbook computes contact details with a
formula, so the cell is `t="str"` with a cached string result — and calamine 0.26
tried a float parse on that string first, which in Rust accepts a leading `+`.
`Utils.sendCode` would then have texted a number with no country code. Fixed by moving
to **calamine 0.36**, with two tests pinning it.

**`zip`'s default features cannot compile for a browser.** They pull bzip2, zstd and
lzma — C libraries with no wasm32 target — so nothing in this module could ever have
run in a browser, which is the whole point of the work. Now `default-features = false`
with `deflate` only, the only method an import archive uses.

Both were found by having two implementations to compare, not by reading the code.

## The CLI

`--check-only` reports without writing. `--strict` refuses to write when there are
warnings, which is what CI wants. `--base-export` takes a `.json` or an export `.zip`.
`--templates-dir` overrides any of the eight entity templates and says which it took;
a `.hbs` whose name is not one of them is reported rather than ignored, because that
is a typo. `--auth-preset none` builds without configuring authentication — which the
SEIU workbook needs, since it declares SAML and leaves the IdP metadata URL blank
pending the client's identity provider.

`--out` and `--created-at` keep the names the Python used: their documentation is
already written and already read. `--validate-only` becomes `--check-only`, matching
the importer's own `check_only` rather than sitting one word away from it.

Validation runs twice, and they are different questions. The builder reports what is
wrong with the **workbook**, in sheet-and-row terms an author can act on.
`validate()` then reports what would be wrong with the **bundle** — the same check
windmill runs before importing.

## The browser surface

`checkBundle` **ships in the existing WASM package with no workflow change**:
`election_config::wasm` is gated the way `crate::wasm` is, and `build_wasm.yml`
already enables both features it needs. So the admin portal can tell an operator what
is wrong with an export before they upload it, and the answer is the importer's own.

`buildFromWorkbook` and `authPresets` are additionally gated on the xlsx and archive
features, which that build does not enable — so the four existing consumers gain the
validator and carry nothing else. Packaging a build that does enable them lands with
the SPA that needs it.

Plain JS values, not wasm-bindgen classes: a front end holds these in state and hands
them to React, and an opaque handle with a `free()` method is a memory leak waiting
for whoever forgets to call it. A failed build returns its problems rather than
throwing, because a list of problems is something a page can render.

## The browser runs the same fixtures the Rust tests do

`fixtureCases()` hands over the suite from PR 2982 as data, rather than something a
front end reimplements. A page asserting `checkBundle(case.bundle)` matches
`case.expect` is checking that the browser and the server reach the same verdict —
which is the only thing that makes one validator worth having.

```ts
import {fixtureCases, checkBundle} from "sequent-core";

for (const c of fixtureCases()) {
    const report = checkBundle(JSON.stringify(c.bundle));
    // c.expect.errors / c.expect.warnings are the codes this must produce
}
```

## Verified

- `cargo test … election_config` — **400 passed**, 0 failed.
- `cargo check` clean for `sequent-core` at every gate, and for `windmill`,
  `harvest` and `step-cli` individually; no warnings from the new code;
  `cargo fmt --check` and `cargo clippy` clean.

  As on PR 2982, this was **false when first written** — an ungated re-export
  meant the crate did not build for the feature sets its consumers declare. It
  is now checked by a six-way `cargo check --lib` matrix in `tests.yml` rather
  than by hand.
- End to end against the real SEIU1000 workbook, as above. The workbook and its output were kept out of the repository and deleted from the build host afterwards.

**Not verified:** the wasm32 cross-compile itself. `ring`'s build script needs a clang
targeting wasm32, which Apple's lacks — the baseline fails identically on this machine
with and without this change, so it is environmental. CI builds that target today, and
the `zip` fix above is precisely what its first run of the config features would
otherwise have hit.

## Reading it

Four commits: the `step-cli` subcommand, the option-name alignment, the WASM surface,
and the fixtures handed to the browser.

There is also a merge commit bringing PR 2982 forward. A repository rule forbids
force-pushing this branch, so the stack is kept current by merging rather than
rebasing; the diff against the base is unaffected.

## Screenshots/videos Before

`<empty>`

```
built seiu1000-leadership-2027
  import this: out/seiu1000-leadership-2027/seiu1000-leadership-2027.zip
    export_election_event-e8e06504-….json
    export_voters-e8e06504-….csv
    export_scheduled_events-e8e06504-….csv
    export_reports-e8e06504-….csv
  beside it, not part of the import:
    out/seiu1000-leadership-2027/admin_users.csv
    out/seiu1000-leadership-2027/export_permissions-….csv
    out/seiu1000-leadership-2027/templates/…
    out/seiu1000-leadership-2027/keycloak_event_realm_patch.json
  warning: admin_users.csv may carry clear-text passwords. It is a secret, not a deliverable.
```

_A screenshot of the imported event follows once this is run against a live
environment._

---

### The census the builder could not read, and the verdict that lied (`EA-F1-001`, `EA-A-003`)

**`voters_sheet` never wrote `area.external_id`.** `build_tables::voter_area_name`
reads it off every voter row and refuses the bundle without it, so **every** plan
carrying a census compiled to nothing — the wizard's ballot preview was blank and its
download produced nothing. The wizard wrote `area_name`, which is what the *finished*
CSV carries; the builder is the thing that turns the id into that name, so it needs
the id. It is now resolved from the voter's area name rather than asking somebody to
type an area twice, because two spellings of one area is a census that imports into
the wrong district.

`the_voters_sheet_matches_what_the_builder_reads` was green throughout: it asserted
every column emitted is one the builder knows, which says nothing about a column the
builder cannot do without. It now checks both directions.
`area.external_id` is deliberately *not* added to `VOTER_LEADING_COLUMNS` — that list
is what the builder derives or reorders on the way out, and this is one it consumes on
the way in.

**`validate_plan` did not check what `compile_plan` requires**, so the Final Review
screen said **Ready to build** for plans that could not be built. Fixed as an
invariant: `a_plan_the_builder_refuses_is_refused_by_the_wizard_first` compiles a
family of plausibly broken plans and asserts the wizard refuses whatever the builder
refuses. It found **four** divergences —

- a voter with no area (accepted here as "the default area", which is not a thing)
- an election with no contests (an error in the builder, a *warning* here)
- two candidates sharing an identifier (**no equivalent check existed at all**, so a
  candidate could vanish from a ballot by being silently replaced)
- `max_votes` above the number of real options

— each now an error with its own message, and identifier uniqueness checked
event-wide, since two contests may not each have a `yes`.

The converse is deliberately not asserted: the wizard may be stricter, and one that
only said what the builder says would be a thinner screen.

**`PublicationPreview::elections()`** also landed here — the beyond half of the
election picker had shipped and this half had not, so the picker had nothing to draw
from. Found while syncing, alongside a **stale index** in the worktree that would have
deleted another agent's voter-messaging work on the next commit; what distinguished
stale from in-progress was that a refactor moving messaging out of `architect.rs` would
have put it in `comms.rs`, and there is no `comms.rs` and no `PlannedMessage` anywhere
on disk.

Note that the native suite cannot see any of the wasm boundary: `wasm.rs` is behind
`#[cfg(all(feature = "wasm", …))]`, so 570 tests passed the whole time the TypeScript
declaration and the Rust struct disagreed. Rebuilt for wasm32 to check, which is the
only thing that would have.

**Verification:** 571 passed. Downstream: 643 `election-architect`, 203
`election-config-ui`, 7/7 Playwright against this core in a real browser.

### DIFF
--- a/.devcontainer/scripts/rebuild-election-config-wasm.sh
+++ b/.devcontainer/scripts/rebuild-election-config-wasm.sh
@@ -51,7 +51,12 @@ node -e '
 ' "${OUT_DIR}" "${PACKAGE_NAME}"
 
 echo "==> Packing..."
-wasm-pack -v pack "${OUT_DIR}"
+# `npm pack` rather than `wasm-pack pack`, which takes the *crate* directory and
+# looks for a `pkg` child inside it — so with a custom --out-dir it goes hunting
+# for pkg-election-config/pkg and fails. `wasm-pack pack` is a wrapper around
+# `npm pack` in the output directory, which is exactly this, and it drops the
+# tarball in the directory it runs from.
+(cd "${OUT_DIR}" && npm pack)
 
 echo
 echo "==> Built ${OUT_DIR}/${PACKAGE_NAME}-0.1.0.tgz"
--- a/.devcontainer/scripts/rebuild-election-config-wasm.sh
+++ b/.devcontainer/scripts/rebuild-election-config-wasm.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# SPDX-FileCopyrightText: 2026 Sequent Tech <legal@sequentech.io>
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+
+# Builds the WASM package the election-configuration SPAs use.
+#
+# This is a *second* package from the same crate, and it exists so the four front
+# ends that already vendor sequent-core do not have to carry what they will never
+# use. They build with `wasmtest,default_features`, which gives them the bundle
+# schema and the validator. This one adds election_config_xlsx, _templates and
+# _archive — a spreadsheet parser, a template engine and a zip writer, none of which
+# belong in the voting portal.
+#
+# wasm-pack takes the npm package name from the crate name, so both builds would
+# otherwise produce sequent-core-0.1.0.tgz. The rename below is what keeps them
+# apart; everything else is the same pipeline as rebuild-sequent-core-full.sh.
+#
+# Run from inside packages/sequent-core via nix develop:
+#   cd packages/sequent-core && nix develop --command ../../.devcontainer/scripts/rebuild-election-config-wasm.sh
+
+PACKAGE_NAME="sequent-election-config"
+OUT_DIR="pkg-election-config"
+
+echo "==> Checking versions..."
+rustc --version
+wasm-pack --version
+
+echo "==> Building ${PACKAGE_NAME} WASM..."
+wasm-pack build \
+    --mode no-install \
+    --out-name index \
+    --out-dir "${OUT_DIR}" \
+    --release \
+    --target web \
+    --features=wasmtest,default_features,election_config_xlsx,election_config_templates,election_config_archive
+
+echo "==> Renaming the package so it does not collide with sequent-core..."
+# node rather than sed: package.json is JSON, and a regex over it is how a build
+# script starts corrupting files that happen to contain the same string twice.
+node -e '
+  const fs = require("fs");
+  const path = process.argv[1] + "/package.json";
+  const manifest = JSON.parse(fs.readFileSync(path, "utf8"));
+  manifest.name = process.argv[2];
+  manifest.description =
+    "Reading, building and validating a Sequent election event import, in the browser";
+  fs.writeFileSync(path, JSON.stringify(manifest, null, 2) + "\n");
+' "${OUT_DIR}" "${PACKAGE_NAME}"
+
+echo "==> Packing..."
+wasm-pack -v pack "${OUT_DIR}"
+
+echo
+echo "==> Built ${OUT_DIR}/${PACKAGE_NAME}-0.1.0.tgz"
+echo
+echo "    Vendor it into whichever package consumes it, the way the four front ends"
+echo "    vendor sequent-core, and update that package's lockfile hash."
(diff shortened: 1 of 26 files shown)


edulix:
@coderabbitai review

